PR #8, opened 2020-06-12
BI-3868 Changing govuk-frontend to version 3.6.0

markpit:


markpit:
> Noticed in the cdn there's 3.6 available there, have you considered using that? Not a big deal but it might be better.

Good spot, I'll change to 3.6

PR #9, opened 2020-06-15
Feature/bi 3883 landing page

mattch1:
Just added a route and template as the cdn styling and parent templates are being done on other tickets.

### DIFF
--- a/src/routes/routes.ts
+++ b/src/routes/routes.ts
@@ -3,7 +3,7 @@ import { NextFunction, Request, Response, Router } from "express";
 const router: Router = Router();
 
 const renderTemplate = (template: string) => (req: Request, res: Response, next: NextFunction) => {
-    return res.render(template, { templateName: template });
+  return res.render(template, { templateName: template });
 };
 
 router.get("/", renderTemplate("index"));
--- a/src/routes/routes.ts
+++ b/src/routes/routes.ts
@@ -1,7 +1,11 @@
-import { Router } from "express";
+import { NextFunction, Request, Response, Router } from "express";
 
 const router: Router = Router();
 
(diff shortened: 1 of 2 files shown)


views/index.html
```diff
@@ -0,0 +1,38 @@
...
+            Object to a company being struck off
+        </h1>
+
+        <p>Use this service to tell us why a limited company should not be removed from the companies register.
+           You can only do this after the notice has been posted in The Gazette.</p>
+        <p>Companies House must receive the objection at least 2 weeks before the notice expiry date.</p>
+
+        <p>This service takes around 15 minutes.</p>
+
+        <a href="/sign-in" role="button" draggable="false" class="govuk-button govuk-button--start govuk-!-margin-top-2 govuk-!-margin-bottom-8" data-module="govuk-button">
```
ICoxCH, line 19: The sign-in page is going to be achieved via an interceptor, so you'll just need to put the next page in here, not a link to sign-in

mattch1: Using link to company number page instead.

src/routes/routes.ts
```diff
@@ -1,7 +1,11 @@
-import { Router } from "express";
+import { NextFunction, Request, Response, Router } from "express";
 
 const router: Router = Router();
 
-router.get("/", (req, response) => response.send("This is the Strike Off Objections app"));
+const renderTemplate = (template: string) => (req: Request, res: Response, next: NextFunction) => {
+    return res.render(template, { templateName: template });
```
mwestacott, line 6: Presume we're still using the same formatting standard as PTF - 2 spaces for indents? If so, the indent here will need amending.

mattch1: Ammended.

PR #10, opened 2020-06-16
Bi 3927 add company number screen

mwestacott:


views/layout.html
```diff
@@ -0,0 +1,124 @@
...
+  <link href="{{CDN_URL}}/stylesheets/services/strike-off-objections/application.css" rel="stylesheet" />
+  <!--<![endif]-->
+
+  {# For Internet Explorer 8, you need to compile specific stylesheet #}
+  {# see https://github.com/alphagov/govuk-frontend/blob/master/docs/installation/supporting-internet-explorer-8.md #}
+  <!--[if IE 8]>
+  <link href="{{CDN_URL}}/stylesheets/govuk-frontend/v3.5.0/govuk-frontend-ie8-3.5.0.min.css" rel="stylesheet" />
+
+  <!-- TODO BI-3878 This CSS must still be added to CDN -->
+  <link href="{{CDN_URL}}/stylesheets/services/strike-off-objections/application.css" rel="stylesheet" />
```
ICoxCH, line 46: This line appears to match the line on 37, can we just have one instance outside the browser if statements? Would that still work?

views/company-number.html
```diff
@@ -0,0 +1,63 @@
...
+          summaryText: "Help with company number",
+          html: '<p>Use the <a target="blank" href="https://beta.companieshouse.gov.uk/">Companies House service (opens in a new tab)</a> to find it.</p>'
+         }) }}
+        {{ govukButton({
+          text: "Continue"
+        }) }}
+      </form>
+    </div>
+  </div>
+{% endblock %}
```
mattch1, line 63: line missing

views/company-number.html
```diff
@@ -0,0 +1,63 @@
...
+            },
+            classes: "govuk-input--company-number",
+            id: "company-number",
+            name: "company-number",
+            errorMessage: {
+              text: "Enter the company number"
+            }
+          }) }}
+        {% else %}
+          {{ govukInput({
```
markpit, line 39: can you look at PTF company-number.html. We can set the error message in it's own variable and this will remove the govukInput duplication

views/layout.html
mwestacott: Not sure. We had the same with PTF and it might be needed for special handling in different browsers. I'll check with @SVarallo though.

SVarallo: I don't see why it wouldn't work to have the if just contain the govuk import and have the application styles outside it.

mwestacott: But @ICoxCH 's suggestion is to have the govuk import outside of the if statement. But doing this would mean it would also apply for older browsers (see block below - [if lt IE 9]) when the import shouldn't be done.

My opinion right now, given as developers we're not actually testing these changes with different browsers and I can't be sure of what the change might do, is that we leave this as is, exactly the same as we had for PTF.

markpit: for IE8 there should be an application-ie8.css.
See govuk-frontend/govuk/all-ie8.scss
Also see the application-ie8.scss in extensions-web

SVarallo: I don't think that's required as govuk-frontend-ie8-3.5.0.min.css covers that. In ptf an application-ie8.css wasn't required.

markpit: in which case we should use govuk-frontend-ie8-3.6.0.min.css as we've moved to 3.6.0 now

views/company-number.html
mwestacott: Added.

views/company-number.html
mwestacott: Done. Although note that error handling and use of this variable will come with implementation of a later story.

views/layout.html
mwestacott: Done. Now using the v3.6.0 css and page rendering correctly.

views/layout.html
```diff
@@ -0,0 +1,124 @@
+{#- We can't mount GOV.UK Frontend's assets at root as it's done automatically by the extensions framework. -#}
```
markpit, line 1: comment references extensions - not sure what this means so remove it?

mwestacott: Will do. It was there in the prototype code.

mwestacott: Didn't want to remove the comment completely - sounds like it would mean something to someone - so just generalised it.

views/layout.html
```diff
@@ -0,0 +1,124 @@
...
+    },
+    html: 'This is a new service - Help us improve it by providing <a class="govuk-link" href="https://www.research.net/" target="_blank">feedback <span class="govuk-visually-hidden">(This link opens in a new window)</span></a>'
+  }) }}
+  {% block backLink %}
+  {% endblock %}
+{% endblock %}
+
+{% set mainClasses = mainClasses | default("govuk-main-wrapper--auto-spacing") %}
+
+{% if useAutoStoreData %}
```
markpit, line 85: this autostoredata block isn't needed, looks like a prototype specific thing

mwestacott: True. I spotted this but left it, as that's what the screenshot on the story has.

views/layout.html
```diff
@@ -0,0 +1,124 @@
+{#- We can't mount GOV.UK Frontend's assets at root as it's done automatically by the application. -#}
+{%- set assetPath = '/govuk/assets' -%}
```
markpit, line 2: do we need the assets? our previous 2 apps didn't have it, unless this is necessary for 3.6.0. I would try deleting it and see if it runs ok.

mwestacott: Will try.

views/layout.html
```diff
@@ -0,0 +1,124 @@
...
+
+{% extends "govuk/template.njk" %}
+
+{% from "govuk/components/accordion/macro.njk" import govukAccordion %}
```
markpit, line 6: i would also remove the imports we don't need, we can add more when we need to. Sorry!

mwestacott: How do I know which ones are and aren't needed? I've lifted this from the prototype, so strikes me as easier to just leave this as is, or, as more prototype screens are added, developers will have to come back to layout.html, work out what imports are missing and add them in.

views/layout.html
mwestacott: Done.

views/layout.html
markpit: you can tell from what is used in each page, govUkInput etc.. We can leave for another PR. It's best to keep to a minimum or we increase client load times if code that is not needed is loaded

PR #11, opened 2020-06-17
Feature/bi 3883 integrate landing page

mattch1:
Unit testing and add layout extension to the landing page. Some font styling issues.

### DIFF
--- a/test/routes/routes.spec.unit.ts
+++ b/test/routes/routes.spec.unit.ts
@@ -4,10 +4,18 @@ import { COOKIE_NAME } from "../../src/properties";
 
 describe("Basic URL Tests", () => {
 
-  it("should find start page", async () => {
+  it("should find start page with cookie", async () => {
     const response = await request(app)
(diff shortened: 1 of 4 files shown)


test/routes/routes.spec.unit.ts
```diff
@@ -1,8 +1,18 @@
 import * as request from "supertest";
 import app from "../../src/app";
+import { COOKIE_NAME } from "../../src/properties";
 
 describe("Basic URL Tests", () => {
 
+  it("should find start page", async () => {
+    const response = await request(app)
+        .get("/strike-off-objections")
+        .set("Cookie", [`${COOKIE_NAME}=123`]);
```
markpit, line 10: what happens if the user is new and doesn't have a cookie? we should test that as well.

mattch1: As it is the landing page the result should be the same. Will add a test to show this.

PR #13, opened 2020-06-17
Feature/bi 3927 remove unused nunjucks imports prototype label

markpit:


views/layout.html
```diff
@@ -79,25 +60,6 @@
 
 {% set mainClasses = mainClasses | default("govuk-main-wrapper--auto-spacing") %}
 
-{% if useAutoStoreData %}
```
mwestacott, line 82: Removing this will mean the implementation of https://companieshouse.atlassian.net/browse/BI-3927 won't match the attached screenshot. We were strictly adhering to screenshots I thought to avoid confusion, so we either need the screenshot changed (on the landing page story too) or this should stay.

markpit: we'll be here forever then

mwestacott: Well we are until Jan 2021.

markpit: I've updated the screenshot, Sarah won't know how to remove the prototype links from the screenshot so our screens will never match the prototype screenshot.

mwestacott: Thanks for doing that. PTF story screens matched so maybe the prototype screens need updating? For example, see screenshots on LFA-1454. Anyway, all good. Approving.

PR #16, opened 2020-06-18
BI-4007 service offline landing page

mattch1:
Similar to promise to file - created middleware to load an alternative template if the flag is set.

### DIFF
--- a/src/feature.flag.ts
+++ b/src/feature.flag.ts
@@ -2,7 +2,7 @@
  * Feature flags will be determined by environment variables and all environment variables in nodejs are
  * either string or undefined. This function will ensure that 'false', '0', 'off' etc remain falsy
  */
-export default (flag: string | undefined): boolean => {
+export const isActiveFeature = (flag: string | undefined): boolean => {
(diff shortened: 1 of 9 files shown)


src/feature.flag.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Feature flags will be determined by environment variables and all environment variables in nodejs are
+ * either string or undefined. This function will ensure that 'false', '0', 'off' etc remain falsy
+ */
+export default (flag: string | undefined): boolean => {
```
markpit, line 5: I was thinking as an improvement, we should give this function a name as when we import it we have to specify a name each time eg. isActiveFeature

markpit:
looks ok but concourse check has failed

mattch1:
I'm not sure how sonar is find this error there is no duplication -seem like a false alarm.

markpit:
> I'm not sure how sonar is finding this error there is no duplication -seems like fake news.

I think you need to merge the latest master into your branch, looks like a possible merge conflict

PR #18, opened 2020-06-18
BI-4030 Adding logger functionality

markpit:
Adding the ch-structured-logging-node logger
needed to add esModuleInterop to the tsconfig which in turn changed the way the express import is made in app.ts. The tsconfig flag is to allow mixing of js and typescript exports from modules (or something like that!)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   "dependencies": {
     "bufferutil": "^4.0.1",
     "canvas": "^2.6.1",
+    "ch-logging": "git+ssh://git@github.com/companieshouse/ch-structured-logging-node.git#a3e0dcc57f91f63803e02ceef2cc5e4038de9a19",
     "express": "~4.16.1",
     "govuk-frontend": "^3.6.0",
     "nunjucks": "^3.2.1",
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,0 +1,9 @@
+import { createLogger } from "ch-logging";
+import { APP_NAME } from "./app";
+import { LOG_LEVEL } from "./properties";
+
+const logger = createLogger(APP_NAME);
+
+logger.info(`env.LOG_LEVEL set to ${LOG_LEVEL}`);
+
+export default logger;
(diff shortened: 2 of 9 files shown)


ICoxCH:
Oops, was a bit quick on the approval, its not a fan of the unit tests anymore

markpit:
> Oops, was a bit quick on the approval, its not a fan of the unit tests anymore

It doesn't like some 'import * as ' imports now, due to that esModuleInterop flag. We have to change the to import x from "blah". Should be fixed now.

package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "bufferutil": "^4.0.1",
     "canvas": "^2.6.1",
+    "ch-logging": "git+ssh://git@github.com/companieshouse/ch-structured-logging-node.git#a3e0dcc57f91f63803e02ceef2cc5e4038de9a19",
```
mwestacott, line 19: Presume this repo will be properly released via Concourse with a release version at some point?

src/logger.ts
```diff
@@ -0,0 +1,9 @@
+import { createLogger } from "ch-logging";
+import { APP_NAME } from "./app";
+import { LOG_LEVEL } from "./properties";
+
+const logger = createLogger(APP_NAME);
+
+logger.info(`env.LOG_LEVEL set to ${LOG_LEVEL}`);
```
mwestacott, line 7: Guessing if log level is ever set lower (or is it higher?) than INFO - e.g. ERROR - this line of output will never appear.

package.json
markpit: I guess so, it's not our repo

src/logger.ts
markpit: yes, info is highest we can go without using error

markpit: there is also WARN but doesn't seem right to use that here

markpit: forget that, warn is not implemented in this logger library

mwestacott: I did wonder about whether this line should be output as a console.log message if the intention is to know what the log level is before any logging is done by the framework. Not sure that's right either though...

markpit: that should work, i can change it

PR #20, opened 2020-06-19
BI-4030 Adding company profile service

markpit:
Adding company profile service to get the company profile from CHS
Also added the api-enumerations submodule for translating some company profile values

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,3 @@ node_modules/
 *.code-workspace
 .DS_Store
 coverage
-/api-enumerations/
-/dist/
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ clean:
 	rm -f ./build.log
 
 .PHONY: build
-build:
+build: update_submodules
 	npm i
 	npm run build
 
@@ -44,8 +44,8 @@ endif
 	$(info Packaging version: $(version))
 	$(eval tmpdir := $(shell mktemp -d build-XXXXXXXXXX))
 	cp -r ./dist/* $(tmpdir)
-	#mkdir $(tmpdir)/api-enumerations
-	#cp ./api-enumerations/*.yml $(tmpdir)/api-enumerations
+	mkdir $(tmpdir)/api-enumerations
+	cp ./api-enumerations/*.yml $(tmpdir)/api-enumerations
 	cp -r ./package.json $(tmpdir)
 	cp -r ./package-lock.json $(tmpdir)
 	cp ./start.sh $(tmpdir)
@@ -57,3 +57,7 @@ endif
 
 .PHONY: dist
 dist: lint test-unit clean package
+
+.PHONY: update_submodules
+update_submodules:
+	test -f ./api-enumerations/constants.yml || git submodule update --init --recursive -- api-enumerations
--- a/src/model/objection.company.profile.ts
+++ b/src/model/objection.company.profile.ts
@@ -1,0 +1,13 @@
+export interface ObjectionCompanyProfile {
+  companyName: string;
+  companyNumber: string;
+  companyStatus: string;
+  companyType: string;
+  address: {
+    line_1: string;
+    line_2: string;
+    postCode: string;
+  };
+  jurisdiction: string;
+  incorporationDate: string;
+}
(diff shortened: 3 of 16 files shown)


.gitignore
```diff
@@ -7,5 +7,3 @@ node_modules/
 *.code-workspace
 .DS_Store
 coverage
-/api-enumerations/
-/dist/
```
markpit, line 11: dist already in here and api-enumerations just gets added as a submodule - git should handle it

test/services/company.profile.service.spec.unit.ts
```diff
@@ -0,0 +1,98 @@
...
+    try {
+        await getCompanyProfile("", ACCESS_TOKEN);
+    } catch (e) {
+        expect(e.status).toEqual(400);
+    }
+    expect.assertions(1);
+  });
+});
+
+const dummySDKResponse: Resource<CompanyProfile> = {
```
mattch1, line 42: Could this go into a separate util file? It looks like something we may reuse.

test/utils/date.formatter.spec.unit.ts
```diff
@@ -0,0 +1,14 @@
+import { formatDateForDisplay } from "../../src/utils/date.formatter";
+
+describe("date formatter tests", () => {
+
+  it("should return a formatted date for displaying when given a date", () => {
+    const formattedDate: string = formatDateForDisplay("1906-11-03");
```
mattch1, line 6: We could add a test for other formats "dd/mm/yyyy" for example.

test/utils/date.formatter.spec.unit.ts
```diff
@@ -0,0 +1,14 @@
...
+    expect(formattedDate).toEqual("3 November 1906");
+  });
+
+  it("should return 'Invalid date' when given empty string", () => {
+    const formattedDate: string = formatDateForDisplay("");
```
mattch1, line 11: Could add a test for undefined as well.

Makefile
```diff
@@ -16,7 +16,7 @@ clean:
 	rm -f ./build.log
 
 .PHONY: build
-build:
+build: update_submodules
```
mwestacott, line 19: Minor, but looks like this should be on a new line for consistent formatting.

src/model/objection.company.profile.ts
```diff
@@ -0,0 +1,13 @@
...
+  companyName: string;
+  companyNumber: string;
+  companyStatus: string;
+  companyType: string;
+  address: {
+    line_1: string;
+    line_2: string;
+    postCode: string;
+  };
+  jurisdiction: string;
```
mwestacott, line 11: I can't see where jurisdiction is used (at least at present). Shouldn't we include this later when we know it's needed?

Makefile
markpit: https://www.gnu.org/software/make/manual/html_node/Rule-Syntax.html
If it's a prerequisite target then it has to go on same line, won't run otherwise.

src/model/objection.company.profile.ts
markpit: ok

test/services/company.profile.service.spec.unit.ts
markpit: I want to avoid using shared util files for test data as each test should be independent from each other and have it's own test data. Besides, this is the only file that will get company info from the SDK.

Makefile
mwestacott: Ah ok 👍

test/utils/date.formatter.spec.unit.ts
markpit: now updated with test for undefined

test/utils/date.formatter.spec.unit.ts
markpit: added tests for some other formats, there are probably lots more formats but I've now enable strict mode on the date formatter so any format that isn't the CHS format YYYY-MM-DD should return invalid date.

test/utils/date.formatter.spec.unit.ts
mattch1: Seen this and approved the PR

test/utils/date.formatter.spec.unit.ts
mattch1: Seen these and approved the PR

PR #21, opened 2020-06-22
Updated text on start page and fixed links in current pages

SVarallo:
Jira ticket at: https://companieshouse.atlassian.net/browse/BI-3883


mwestacott:
I think there might be a failing test due to the screen text changes.

PR #22, opened 2020-06-22
BI-3925 Add authentication and session middleware

markpit:
Adding CH session and authentication middleware


src/middleware/authentication.middleware.ts
```diff
@@ -0,0 +1,11 @@
+import { authMiddleware } from "web-security-node";
+import * as pageURLs from "../model/page.urls";
+
+const authMiddlewareConfig = {
+  accountWebUrl: "",
+  returnUrl: `${pageURLs.STRIKE_OFF_OBJECTIONS}${pageURLs.COMPANY_NUMBER}`,
+};
+
+const signInMiddleware = authMiddleware(authMiddlewareConfig);
```
mwestacott, line 9: Perhaps call this authenticationMiddleware and export/import it as this. Then consistent with sessionMiddleware, below.

src/app.ts
```diff
@@ -1,9 +1,13 @@
+import cookieParser from "cookie-parser";
 import express from "express";
 import * as nunjucks from "nunjucks";
 import * as path from "path";
 import { checkServiceAvailability } from "./availability/middleware/service.availability";
 import { APP_NAME } from "./constants";
 import logger from "./logger";
+import signInMiddleware from "./middleware/authentication.middleware";
```
mwestacott, line 8: See suggested import name change below.

src/middleware/authentication.middleware.ts
markpit: will do

src/app.ts
markpit: done

src/middleware/authentication.middleware.ts
markpit: done

PR #24, opened 2020-06-24
Feature/bi 4032 populate company details

mattch1:
Some unit tests to be updated but the other code can be reviewed.


src/controllers/company.number.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+import logger from "../utils/logger";
+
+/**
+ * POST controller for company number screen
+ * @param req
+ * @param res
+ * @param next
+ */
+const route = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+    logger.debug("Attempt to retrieve and show the company details");
```
markpit, line 16: as this is logging that we've entered the method, you can use debugRequest which will also log out the url path
eg. logger.debugRequest(req, "Attempt to retrieve and show the company details");

src/controllers/company.number.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+    const company: ObjectionCompanyProfile = await getCompanyProfile(companyNumber, token);
+
+    if (session) {
+        createObjectionsSession(session);
+        addToObjectionsSession(session, "company_data", company);
+    }
+    return res.redirect(OBJECTIONS_CONFIRM_COMPANY);
+};
+
+const getValidToken = (session: Session, req: Request): string | undefined => {
```
markpit, line 32: I would move this method into the new objections.session.service. I can see it being needed by other controllers in future and feels like it is a better 'fit' for this method.

src/controllers/company.number.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    const companyNumber: string = req.body.companyNumber;
+    const session: Session = req.session as Session;
+    logger.info(`Retrieving company profile for company number ${companyNumber}`);
+    const token: string = getValidToken(session, req) as string;
+
+    const company: ObjectionCompanyProfile = await getCompanyProfile(companyNumber, token);
+
+    if (session) {
+        createObjectionsSession(session);
```
markpit, line 26: createObjectionsSession(session); - This is ok for the time being but I think for another story / PR it would be better to create the objections session in middleware, have middleware that calls createObjectionsSession after the sessionMiddleware has run or even as part of the existing session middleware (if objections session doesn't exist). Requires some thought about how we wipe data from the session perhaps, one for another story I guess.

mattch1: Yes this overlaps with further work needed for sessions handling.

src/controllers/company.number.controller.ts
mattch1: Moved and added unit tests.

src/controllers/company.number.controller.ts
mattch1: updated 2 logging methods with this.

src/controllers/company.number.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+        addToObjectionsSession(session, "company_data", company);
```
mwestacott, line 27: The string "company_data" appears here and also in a class below (possible other places). We defined a constant for this in PTF. Suggest doing the same here.

src/services/objections.session.service.ts
```diff
@@ -12,3 +14,13 @@ export const addToObjectionsSession = (session: Session, key: string, value: any
 export const getValueFromObjectionsSession = (session: Session, key: string) => {
     return session.data[OBJECTIONS_SESSION][key];
 };
+
+export const getValidToken = (session: Session): string | undefined => {
```
markpit, line 18: sorry, should have spotted it earlier, could you rename to getValidAccessToken? makes it clearer as to what the token is for and matches up with the messages being logged.

src/controllers/confirm.company.controller.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * GET controller for check company details screen
+ * @param req
+ * @param res
+ * @param next
+ */
+
+export const route = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+    if (req.session && req.session.data) {
+        const company: string = getValueFromObjectionsSession(req.session, "company_data");
```
mwestacott, line 15: I think that rather than 'string' you can use 'ObjectionCompanyProfile' as the type here?

mattch1: I don't follow - the string is the key.

src/controllers/company.number.controller.ts
mattch1: Can make this a constant.

src/controllers/confirm.company.controller.ts
mwestacott: Sorry, I mean 'const company: ObjectionCompanyProfile...'. Will that work? Stronger typing...

mattch1: Refactored this.

src/services/objections.session.service.ts
mattch1: Updated the name

src/controllers/confirm.company.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const route = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+    if (req.session && req.session.data) {
+        const company: string = getValueFromObjectionsSession(req.session, OBJECTIONS_COMPANY_PROFILE);
+        return res.render(Templates.CONFIRM_COMPANY, {
+            company,
+            templateName: Templates.CONFIRM_COMPANY,
+        });
+    } else {
+        logger.error("Error displaying company data");
```
markpit, line 22: could you use logger.errorRequest(req, ...    here, it will hopefully allow us to work out what function the error occurred in. Could you also put more info in the message, just thinking if this happens in live it won't help us much. Something like "Confirm Company - no company data found in session"

test/controller/confirm.company.spec.unit.ts
```diff
@@ -0,0 +1,61 @@
...
+        expect(response.text).toContain("00006400");
+        expect(response.text).toContain("Active");
+        expect(response.text).toContain("26 June 1872");
+        expect(response.text).toContain("limited");
+        expect(response.text).toContain("line1");
+        expect(response.text).toContain("line2");
+        expect(response.text).toContain("post code");
+    });
+});
+
```
markpit, line 47: there is a scenario where an error is logged if company data is missing from session. We should have a test to check that it happens correctly. We should be able to mock the logger and check that it is called with the correct error message.

src/constants.ts
```diff
@@ -1 +1,3 @@
 export const APP_NAME = "strike-off-objections-web";
+export const OBJECTIONS_SESSION = "objections_session";
```
markpit, line 2: can you name it strike_off_objections_session as this means it won't get confused if other objections services are created in the future within Redis
eg . export const OBJECTIONS_SESSION_NAME = "strike_off_objections_session";

src/constants.ts
```diff
@@ -1 +1,3 @@
...
+export const OBJECTIONS_COMPANY_PROFILE = "objections_company_profile";
```
markpit, line 3: we might have a few session constants so we can group them by prefixing SESSION_
also we know it's for objections so we can remove that bit 
eg. 
export const SESSION_COMPANY_PROFILE = "objections_company_profile"

test/controller/confirm.company.spec.unit.ts
mattch1: Added TODO comment for this work.

src/controllers/confirm.company.controller.ts
mattch1: Updated error logging

src/constants.ts
mattch1: Changed the name

src/constants.ts
mattch1: Updated.

src/model/page.urls.ts
```diff
@@ -5,7 +5,11 @@ import { Templates } from "./template.paths";
  */
 
 const SEPARATOR: string = "/";
+export const OBJECTIONS: string = SEPARATOR + "strike-off-objections";
```
mwestacott, line 8: We don't need this new constant definition - you can just use STRIKE_OFF_OBJECTIONS defined below.

src/model/page.urls.ts
```diff
@@ -5,7 +5,11 @@ import { Templates } from "./template.paths";
...
 
 export const STRIKE_OFF_OBJECTIONS: string = SEPARATOR + "strike-off-objections";
 export const COMPANY_NUMBER: string = SEPARATOR + Templates.COMPANY_NUMBER;
 export const CONFIRM_COMPANY: string = SEPARATOR + Templates.CONFIRM_COMPANY;
+
+export const OBJECTIONS_COMPANY_NUMBER: string = OBJECTIONS + COMPANY_NUMBER;
```
mwestacott, line 14: A bit confused why we need these as well as the two definitions above. These two have the full path the ones above don't.

mwestacott: Ah just seen the same was done in PTF so probably needed (for redirects). Worth lifting the comments in the PTF page.urls file and copying here I think - they explain why there is partial duplication...

mattch1: We don't always need the full path but sometimes we do.

test/controller/company.number.controller.spec.unit.ts
```diff
@@ -0,0 +1,102 @@
...
+        const mockValidAccessToken = getValidAccessToken as jest.Mock;
+
+        beforeEach(() => {
+            mockValidAccessToken.mockReset();
+        });
+
+        mockValidAccessToken.mockReset();
+        mockValidAccessToken.mockImplementation(() => ACCESS_TOKEN );
+
+        mockCompanyProfile.mockResolvedValueOnce(dummyCompanyProfile);
```
markpit, line 74: mockCompanyProfile.mockResolvedValueOnce(dummyCompanyProfile);  is returning the function rather than executing the function, so in the controller, a function is being added to the session instead of the company profile object.
if you change it to 
mockCompanyProfile.mockResolvedValueOnce(dummyCompanyProfile());
with the extra () - it will execute teh function and return teh company profile.  then you can add an expect(company name).toEqual("blah).  but either take the apostrophe out of the name or in teh expect translate the apostrophe to teh html code for apostrophe.

markpit: or change getDummyCompanProfile to be just an object instead of a function - see confirm.company.controller test - either way is ok.

mattch1: changed to dummyCompanyProfile

test/controller/company.number.controller.spec.unit.ts
```diff
@@ -0,0 +1,100 @@
...
+
+        const response = await request(app)
+            .post(OBJECTIONS_COMPANY_NUMBER)
+            .set("Accept", "application/json")
+            .set("Referer", "/")
+            .set("Cookie", [`${COOKIE_NAME}=123`])
+            .send({companyNumber: COMPANY_NUMBER});
+
+        expect(response.header.location).toEqual(OBJECTIONS_CONFIRM_COMPANY);
```
markpit, line 83: I would also add a couple of expects to check company info is on the page correctly like you've done for confirm company

mattch1: This is why I didn't add any expects because we are covering this with the other test fopr the company confirmation screen.

src/model/page.urls.ts
mattch1: Using this.

test/controller/company.number.controller.spec.unit.ts
markpit: sorry I got it wrong, the company details aren't shown on that screen, I was confused with the confirm screen. Nothing to do here.

PR #25, opened 2020-06-24
BI-4032 Fixes for confirm company test

markpit:
test was hanging due to 2 things:
a) needed to mock redis out with jest.mock("ioredis");
b) the route function was falling into the else block as the session had no data and it needed to return something after it had logged the error otherwise it just hangs. I've added return next() to get the test to pass but unsure if this will work in real life (haven't run it). We need to implement the generic error screen so we can call next(new Error("blah"));  

added the data block to the session in the mock implementation to get past the 'if' in the controller. I'll look into whether we can make that mock importable, but for the time being just do similar in each test.

tidied up a few things, the getDummyProfile function in the test was returning a function which was passed into the template so fixed that.
Removed async off route as I think we only need to use async if we use await inside.
VSCode highlighted the supertest types so added those.
Changed the route from being exported as an array, might need to change it back in future if we want to export a chain of functions.
Made the ObjectionsCompanyProfile interface the default export as it will be the only thing in that file. We can then remove the {} from the import.

### DIFF
--- a/src/routes/routes.ts
+++ b/src/routes/routes.ts
@@ -19,6 +19,6 @@ router.get("/", renderTemplate(Templates.INDEX));
 
 router.get(pageURLs.COMPANY_NUMBER, renderTemplate(Templates.COMPANY_NUMBER));
 router.post(pageURLs.COMPANY_NUMBER, ...companyNumberRoute);
-router.get(pageURLs.CONFIRM_COMPANY, ...confirmCompanyRoute);
+router.get(pageURLs.CONFIRM_COMPANY, confirmCompanyRoute);
 
 export default router;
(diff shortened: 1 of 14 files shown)


src/routes/routes.ts
```diff
@@ -19,6 +19,6 @@ router.get("/", renderTemplate(Templates.INDEX));
 
 router.get(pageURLs.COMPANY_NUMBER, renderTemplate(Templates.COMPANY_NUMBER));
 router.post(pageURLs.COMPANY_NUMBER, ...companyNumberRoute);
```
mwestacott, line 21: Just wondering if we should (can) change that .post function above to not use the ellipses '...'?

markpit: yes we can, i've updated it now

PR #27, opened 2020-06-26
BI-4115 New service and stub for creating a new objection

mwestacott:


src/services/objections.api.service.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Create a new objection for the given company.
+ *
+ * @param companyNumber the company number
+ * @param token the bearer security token to use to call the api
+ *
+ * @returns the id of the newly created objection
+ */
+export const createNewObjection = async (companyNumber: string, token: string): Promise<NewObjection> => {
```
markpit, line 12: you can remove async/promise if you aren't making any await calls

src/services/objections.api.service.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  logger.debug(`Creating a new objection for company number ${companyNumber}`);
+
+  // TODO Call the actual Objections API (via the SDK Node API?) when end-point is implemented. Covered by
+  //      JIRA sub-task BI-4121
+
+  return {
```
markpit, line 19: i'd prefer to return just the id string, that way the client won't have to know about the structure of the object being returned. The model should be in the SDK (when we write it) so that can probably go.

markpit: ignore that, we do need to return a new Objection. ignore above.

mwestacott: After further discussion agreed to just return id as a string. Objection object will come later, when we integrate the call to GET an objection.

PR #30, opened 2020-06-29
BI-4123 Adding PATCH stub to objections.api.service

markpit:


src/services/objections.api.service.ts
```diff
@@ -1,5 +1,22 @@
 import logger from "../utils/logger";
 
+/**
+ * State of Objection
+ */
+export enum ObjectionStatus {
```
markpit, line 6: I'm keeping everything in the same module for the time being, should be easier to migrate to the sdk in future.

src/services/objections.api.service.ts
```diff
@@ -1,5 +1,22 @@
 import logger from "../utils/logger";
 
+/**
```
markpit, line 3: I'm keeping everything in the same module for the time being, should be easier to migrate to the sdk in future.

PR #32, opened 2020-06-30
BI-3927 Added route to enter-information page from confirm-company

SVarallo:


views/confirm-company.html
```diff
@@ -78,11 +78,11 @@ <h2 class="govuk-heading-l">{{company.companyName}}</h2>
         ]
       }) }}
 
-      <form action="" method="post">
+      <form method="get" action="/strike-off-objections/enter-information">
```
markpit, line 81: forms should be submitted via a POST

SVarallo: This is how it's done in PTF and as far as the story goes we're not posting anything

mwestacott: @markpit might be thinking about the PATCH call to the API as we move to the next screen. I'm happy to consider making a change to POST when I integrate the call to the web stub, so this looks ok for now.

markpit: POST is better for security as GET translates form-data into url attributes viewable in teh browser. https://www.w3schools.com/tags/att_form_method.asp

mwestacott: We're not sending any form data at this point though. Are we?

markpit: the GET url places the attributes marker on '?' on the end of the url - doesn't look good if there are no attributes.
http://web.chs-dev.internal:4000/strike-off-objections/enter-information?

PR #33, opened 2020-06-30
BI-4115 Call service to create the new objection

mwestacott:


test/controller/enter.information.controller.spec.unit.ts
```diff
@@ -0,0 +1,70 @@
...
+
+describe("enter information tests", () => {
+
+    it("should call the API to create a new objection and then render the page", async () => {
+
+        mockGetObjectionSessionValue.mockReset();
+        mockGetObjectionSessionValue.mockImplementation(() => dummyCompanyProfile);
+
+        mockCreateNewObjection.mockReset();
+        mockCreateNewObjection.mockImplementation(() => "123456");
```
markpit, line 44: cold you test if the new objection id is in the session? you should be able to see it in req.session.data["OBJECTIONS_SESSION"] or whatever we've called it.

mwestacott: Session is mocked, so I've checked that the correct, new objection id is passed into the function that saves it in the session.

PR #34, opened 2020-06-30
BI-4121 make call to objections api to create new objection

markpit:
Implemented the create new objection api call.

In order to make it easier to migrate to the real SDK, i've created our own internal sdk module under src/modules/sdk/objections. Anything that should go into the real SDk can now go into that folder.


src/sdk/objections/service.ts
```diff
@@ -0,0 +1,82 @@
+import { AxiosError, AxiosRequestConfig, AxiosResponse, Method } from "axios";
```
mattch1: I'm not sure about calling this a service if it is caling the api it seems more like a client. Also not so sure about the folder name.

src/sdk/objections/service.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const HTTP_POST: Method = "post";
+
+/**
+ * An axios config with common elements for API calls.
+ * @param {Method} httpMethod the http method to use eg. get, post
+ * @param {string} url of the api endpoint
+ * @param {string} token Bearer token for API call
+ */
+const getAxiosRequestConfig = (httpMethod: Method, url: string, bearerToken: string): AxiosRequestConfig => {
```
mattch1: Could the axios calls go into a separate file?

src/services/objections.api.service.ts
```diff
@@ -1,52 +1,46 @@
-import logger from "../utils/logger";
-
```
mattch1, line 2: This definitely seems more like a client if it calls the sdk, so we could say
client -> sdk proxy service -> api

src/modules/sdk/objections/service.ts
```diff
@@ -0,0 +1,82 @@
...
+    logger.debug(`data returned from axios api call : ${JSON.stringify(axiosResponse.data, null, 2)}`);
+    return axiosResponse;
+  } catch (err) {
+    logger.error(`ERROR calling API ${JSON.stringify(err, null, 2)}`);
+    const axiosError = err as AxiosError;
+    const { response, message } = axiosError;
+    throw {
+      data: response ? response.data.errors : [],
+      message,
+      status: response ? response.status : -1,
```
mwestacott, line 46: The -1 is 'magic'! Could you define a constant explaining what this represents? (probably just a general error?). Maybe the definition should live in the types.ts file?

src/services/objections.api.service.ts
```diff
@@ -1,52 +1,46 @@
...
+ * @param {string} companyNumber the company number
+ * @param {string} token the bearer security token to use to call the api
  *
- * @returns the id of the newly created objection
+ * @returns {string} the id of the newly created objection
+ * @throws {ApiError}
  */
-export const createNewObjection = (companyNumber: string, token: string): string => {
+export const createNewObjection = async (companyNumber: string, token: string): Promise<string> => {
+  logger.info(`Creating objection for company number ${companyNumber}`);
```
mwestacott, line 15: I wonder if we need this class now? If all we have in it are a bunch of one-line functions that simply delegate to the sdk class functions it's arguably a pretty useless additional layer of code.

src/sdk/objections/service.ts
markpit: this new structure is now mirroring the module structure used ch-sdk-node 
https://github.com/companieshouse/ch-sdk-node/tree/master/src/services/company-officers
https://github.com/companieshouse/ch-sdk-node/tree/master/src/services/lfp
https://github.com/companieshouse/ch-sdk-node/tree/master/src/services/order/basket

this will help to minimise changes when we come to migrate.

src/sdk/objections/service.ts
markpit: yes I can look at that

src/services/objections.api.service.ts
markpit: I followed the approach taken by Appeals
https://github.com/companieshouse/lfp-appeals-frontend/blob/master/src/modules/appeals-service/AppealsService.ts
There they use services which the controllers interact with and those services can then call the modules they need to complete the function.
The idea is to have an objections service that the controllers can interact with.
It also provides loose coupling with the sdk, a single point of contact. If all interactions are with the service then it minimises change if the sdk interface changes - we wouldn't have to update several controllers, only the service.

src/modules/sdk/objections/service.ts
markpit: i'll look into that

src/services/objections.api.service.ts
markpit: The idea is to have an objections service that the controllers can interact with.
It also provides loose coupling with the sdk, a single point of contact. If all interactions are with the service then it minimises change if the sdk interface changes - we wouldn't have to update several controllers, only the service.

For example, I've split the patch request into 2 separate functions now, one to update the reason, other to update the status. There will have to be some patch object creation done in here (objections service). This will ensure that the SDK types such as Status types won't be polluting other code outside the service such as the controllers. Only the objections service should be exposed to the SDK code.

src/services/objections.api.service.ts
mattch1: From what you have said this structure seems good - still a bit uneasy about the naming i.e just calling it "service" something more descriptive to reflect what you have said and to distinguish it from the sdk service would be helpful.

src/services/objections.api.service.ts
mwestacott: Yep. Good argument.

src/sdk/objections/service.ts
markpit: extracted axios out into axios.client.ts

src/modules/sdk/objections/service.ts
markpit: I've changed it to a constant STATUS_NO_RESPONSE now and exported it. Didn't think it warranted being a 'type'

src/modules/sdk/objections/service.ts
```diff
@@ -0,0 +1,38 @@
...
+ * @param {string} companyNumber the company number
+ * @param {string} token the bearer security token to use to call the api
+ *
+ * @returns {string} the id of the newly created objection
+ * @throws {ApiError}
+ */
+export const createNewObjection = async (companyNumber: string, token: string): Promise<string> => {
+  logger.info(`Creating a new objection for company number ${companyNumber}`);
+
+  const createNewObjectionUrl = `${INTERNAL_API_URL}/company/${companyNumber}/strike-off-objections`;
```
mwestacott, line 19: Guessing this URL string might become a constant when we add more calls to the API end-points, but not needed yet.

src/services/objections.service.ts
```diff
@@ -0,0 +1,50 @@
...
+  //      sub-task BI-4143
+};
+
+/**
+ * Update objection status to submitted for the given objection ID.
+ *
+ * @param {string} objectionId the id of the objection
+ * @param {string} token the bearer security token to use to call the api
+ */
+export const updateObjectionStatusToSubmitted = (objectionId: string, token: string) => {
```
mwestacott, line 44: How about just 'submitObjection' as the name for this method? The fact that the service issues a patch/update to achieve this is an implementation detail the client doesn't need to know (does it?).

markpit: makes sense, updated

src/modules/sdk/objections/service.ts
markpit: it will be re-used so made into into a constant/function

markpit:
> Still uneasy about the naming similarities between sdk and api objections service but will approve.

ok, I can bring it up in our upload meeting tomorrow and we'll get suggestions from the team.

PR #36, opened 2020-07-03
OBJ-68 stub for adding attachment

mattch1:
Stubbed two service methods api and sdk and stubbed a test

### DIFF
--- a/src/services/objections.service.ts
+++ b/src/services/objections.service.ts
@@ -48,3 +48,13 @@ export const submitObjection = (objectionId: string, token: string) => {
   // TODO Call the Objections SDK. Covered by JIRA
   //      sub-task BI-4143
 };
+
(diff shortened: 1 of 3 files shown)


src/services/objections.service.ts
```diff
@@ -48,3 +48,13 @@ export const submitObjection = (objectionId: string, token: string) => {
...
   //      sub-task BI-4143
 };
+
+export const addAttachment = (companyNumber: string,
+                              token: string,
+                              objectionId: string,
+                              attachment: Buffer,
+                              fileName: string) => {
+
+  logger.info(`Adding attachment to objection`);
```
markpit, line 58: would you be able to add more info into the message?  Like:
Adding attachment ${fileName} to objection ${objectionId}

mattch1: I can do.

PR #37, opened 2020-07-03
obj-43 Add objection session middleware

ICoxCH:


src/controllers/company.number.controller.ts
```diff
@@ -14,20 +17,28 @@ import logger from "../utils/logger";
...
 
-    if (session) {
-        createObjectionsSession(session);
-        addToObjectionsSession(session, SESSION_COMPANY_PROFILE, company);
-    }
+  const company: ObjectionCompanyProfile = await getCompanyProfile(companyNumber, token);
+
+  const objectionsSessionExtraData: ObjectionSessionExtraData = retrieveObjectionsSessionFromSession(session);
+  if (objectionsSessionExtraData) {
+    objectionsSessionExtraData.objections_company_profile = company;
```
markpit, line 37: could you make objections_company_profile a constant, possibly in constants.ts - i did think to create an enum inside constant.ts for the session value constants but I think it should at least be a constant.

ICoxCH: It already is, I forgot to use the constant here though, will update

markpit: thinking some more - would it be better to move this block of code into objections.session.service and have a function like addCompanyProfileToSession

markpit: > It already is, I forgot to use the constant here though, will update

I noticed that it is part of an interface further down the code so it's ok, didn't realise you had defined a new interface

src/controllers/confirm.company.controller.ts
```diff
@@ -13,18 +12,18 @@ import logger from "../utils/logger";
...
+      const company: ObjectionCompanyProfile = retrieveCompanyProfileFromObjectionsSession(req.session);
+      return res.render(Templates.CONFIRM_COMPANY, {
+        company,
+        templateName: Templates.CONFIRM_COMPANY,
+      });
+    } catch (e) {
+      logger.errorRequest(req, "Error retrieving company profile from session");
+      return next(e);
     }
+  }
```
markpit, line 26: I would either add a next(new Error("no session")) as an else block, or remove the if(req.session) from the top and let it fail if no session.

src/middleware/objections.session.middleware.ts
```diff
@@ -0,0 +1,22 @@
+import { Session } from "ch-node-session-handler";
+import { NextFunction, Request, Response } from "express";
+import { OBJECTIONS_SESSION_NAME } from "../constants";
+import ObjectionSessionExtraData from "../model/objection.session.extra.data";
+
+const objectionsSessionMiddleware = (req: Request, res: Response, next: NextFunction) => {
+  const session: Session = req.session as Session;
+  if (session) {
+    const objectionsSession: ObjectionSessionExtraData | undefined = session.getExtraData("objections_session");
```
markpit, line 9: should "objections_session" be replaced with OBJECTIONS_SESSION_NAME?

test/services/objections.session.service.spec.unit.ts
```diff
@@ -1,81 +1,54 @@
...
+          access_token: accessTokenValue,
+        },
+      },
+    };
+    const token: string = retrieveAccessTokenFromSession(session) as string;
+    expect(token).not.toBeUndefined();
+    expect(token).toEqual(accessTokenValue);
+  });
+
+  it("should receive undefined when token is absent", () => {
```
markpit, line 23: not sure if test title is correct, test expects an error to be thrown

src/controllers/company.number.controller.ts
```diff
@@ -14,20 +16,27 @@ import logger from "../utils/logger";
...
+    return next(e);
+  }
 
-    if (session) {
-        createObjectionsSession(session);
-        addToObjectionsSession(session, SESSION_COMPANY_PROFILE, company);
-    }
+  const company: ObjectionCompanyProfile = await getCompanyProfile(companyNumber, token);
+
+  try {
```
mwestacott, line 34: Do we need the two 'try' blocks? Could just have one block I think since error handling is the same here as above. Code then becomes more readable (focussed on the happy path).

src/middleware/objections.session.middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const objectionsSessionMiddleware = (req: Request, res: Response, next: NextFunction) => {
+  const session: Session = req.session as Session;
+  if (session) {
+    const objectionsSession: ObjectionSessionExtraData | undefined = session.getExtraData(OBJECTIONS_SESSION_NAME);
+    if (objectionsSession) {
+      return next();
+    }
+
+    const objSession: ObjectionSessionExtraData = {};
```
mwestacott, line 14: For consistency, should this object be named ObjectionsSessionExtraData?

src/model/objection.session.extra.data.ts
```diff
@@ -0,0 +1,6 @@
+import ObjectionCompanyProfile from "./objection.company.profile";
+
+export default interface ObjectionSessionExtraData {
+  objections_company_profile?: ObjectionCompanyProfile;
```
mwestacott, line 4: Just 'objection_company_profile'? We should really tidy all these artefact names up and decide on a standard.

src/services/objections.session.service.ts
```diff
@@ -1,25 +1,64 @@
...
-export const addToObjectionsSession = (session: Session, key: string, value: any) => {
-    session.data[OBJECTIONS_SESSION_NAME][key] = value;
+export const addCompanyProfileToObjectionsSession = (session: Session, company: ObjectionCompanyProfile): void => {
+  const objectionsExtraData: ObjectionSessionExtraData = retrieveObjectionsSessionFromSession(session);
+  if (objectionsExtraData) {
+    objectionsExtraData[SESSION_COMPANY_PROFILE] = company;
+    return;
+  }
+
+  throw new Error("No objections session extra data to add company too");
```
mwestacott, line 44: to.

PR #38, opened 2020-07-06
OBJ-45 Adding document upload screen

markpit:
Adding the document upload screen (screen only, no functionality behind it yet). Based on similar screen from extensions-web.


src/controllers/enter.information.controller.ts
```diff
@@ -2,12 +2,12 @@ import { Session } from "ch-node-session-handler";
 import { NextFunction, Request, Response } from "express";
 import ObjectionCompanyProfile from "model/objection.company.profile";
 import { SESSION_COMPANY_PROFILE, SESSION_OBJECTION_ID } from "../constants";
+import * as PageUrls from "../model/page.urls";
```
ICoxCH, line 5: would it be better to just import the page urls you need rather than importing them all?

markpit: I can switch them, it read better to me this way.

PR #39, opened 2020-07-06
BI-4143 Remove the stub and call the real PATCH objection API method

mwestacott:


src/controllers/add.documents.controller.ts
```diff
@@ -0,0 +1,28 @@
+import { Session } from "ch-node-session-handler";
+import { NextFunction, Request, Response } from "express";
+import ObjectionCompanyProfile from "model/objection.company.profile";
+import { SESSION_COMPANY_PROFILE, SESSION_OBJECTION_ID } from "../constants";
+import { Templates } from "../model/template.paths";
+import { updateObjectionReason } from "../services/objections.service";
+import { getValidAccessToken, getValueFromObjectionsSession } from "../services/objections.session.service";
+
+const route = async (req: Request, res: Response, next: NextFunction) => {
```
markpit, line 9: the patching should be done in the Post of enter-information screen.

mwestacott: It is - see the routes.ts change below. But do you mean this file needs renaming?

markpit: there should be a function added to the enter.information.controller to do what you've done in the add.document.controller. The patching is for the information/reason and not related to the add.documents page.

mwestacott: Right. Seen your review and now see what you mean. Think I'm going to have to integrate once your PR is merged. Will approve it now.

mwestacott: Updated so that implementation moved to correct enter information post function.

src/modules/sdk/objections/axios.client.ts
```diff
@@ -4,15 +4,18 @@ import { ApiError } from "./types";
...
 
 /**
  * An axios config with common elements for API calls.
  * @param {Method} httpMethod the http method to use eg. get, post
  * @param {string} url of the api endpoint
  * @param {string} bearerToken token for API call
  */
-export const getAxiosRequestConfig = (httpMethod: Method, url: string, bearerToken: string): AxiosRequestConfig => {
+export const getAxiosRequestConfig = (
+    httpMethod: Method, url: string, bearerToken: string, requestData): AxiosRequestConfig => {
```
markpit, line 16: i would either make requestData optional with question mark - requestData? and in turn make data? optional or maybe rename getAxiosRequestConfig to getBaseAxiosRequestConfig (like we did for extensions) and then each function can add to it
eg https://github.com/companieshouse/extensions-web/blob/master/src/client/apiclient.ts#L116

src/modules/sdk/objections/service.ts
```diff
@@ -1,11 +1,14 @@
 import { AxiosRequestConfig } from "axios";
 import logger from "../../../utils/logger";
 import { INTERNAL_API_URL } from "../../../utils/properties";
-import { getAxiosRequestConfig, HTTP_POST, makeAPICall } from "./axios.client";
+import { getAxiosRequestConfig, HTTP_PATCH, HTTP_POST, makeAPICall } from "./axios.client";
 import { ObjectionPatch } from "./types";
 
 const OBJECTIONS_API_URL = (companyNumber: string): string => `${INTERNAL_API_URL}/company/${companyNumber}/strike-off-objections`;
 
+const OBJECTIONS_API_PATCH_URL = (companyNumber: string, objectionId: string): string =>
```
markpit, line 9: you should be able to replace the base part of the url 
```suggestion
const OBJECTIONS_API_PATCH_URL = (companyNumber: string, objectionId: string): string => OBJECTIONS_API_URL(companyNumber) + `/${objectionId}`;
```

src/modules/sdk/objections/axios.client.ts
mwestacott: 👍 Implementing the extensions approach.

PR #41, opened 2020-07-08
OBJ-44 Created error handler controller, unit tests and error page

SVarallo:
Jira ticket at: https://companieshouse.atlassian.net/browse/OBJ-44

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -41,7 +41,7 @@ app.use(`${pageURLs.STRIKE_OFF_OBJECTIONS}/*`, authenticationMiddleware);
 app.use(`${pageURLs.STRIKE_OFF_OBJECTIONS}/*`, objectionSessionMiddleware);
 // apply our default router to /
 app.use(pageURLs.STRIKE_OFF_OBJECTIONS, router);
-app.use(...errorHandler);
+app.use(errorHandler);
 
 logger.info(`************** ${APP_NAME} has started **************`);
 
--- a/src/app.ts
+++ b/src/app.ts
@@ -3,6 +3,7 @@ import express from "express";
 import * as nunjucks from "nunjucks";
 import * as path from "path";
 import { APP_NAME } from "./constants";
+import errorHandler from "./controllers/error.controller";
 import authenticationMiddleware from "./middleware/authentication.middleware";
 import objectionSessionMiddleware from "./middleware/objection.session.middleware";
 import serviceAvailabilityMiddleware from "./middleware/service.availability.middleware";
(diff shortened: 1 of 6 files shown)


views/error.html
```diff
@@ -0,0 +1,23 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+Sorry, there is a problem with the service - GOV.UK
```
markpit, line 4: indentation on Sorry

src/app.ts
```diff
@@ -40,6 +41,7 @@ app.use(`${pageURLs.STRIKE_OFF_OBJECTIONS}/*`, authenticationMiddleware);
 app.use(`${pageURLs.STRIKE_OFF_OBJECTIONS}/*`, objectionSessionMiddleware);
 // apply our default router to /
 app.use(pageURLs.STRIKE_OFF_OBJECTIONS, router);
+app.use(...errorHandler);
```
markpit, line 44: at the moment, the error handler is default exporting an array of one function
export default [errorHandler];
so we should remove the square brackets and then we can remove the three dots here which expands the array (we won't be exporting an array if we remove the square brackets)
In extensions we had an array of 2 functions exported from the controller, the extra one for a 404 page. I can raise the 404 as a separate story.

src/controllers/error.controller.ts
```diff
@@ -0,0 +1,13 @@
+import { NextFunction, Request, Response } from "express";
+import logger from "../utils/logger";
+
+/**
+ * This handler catches any other error/exception thrown within the application. Always keep this as the
+ * last handler in the chain for it to work.
+ */
+const errorHandler = (err, req: Request, res: Response, next: NextFunction) => {
+  logger.error("An error has occurred. Re-routing to the error screen - " + err.stack);
```
markpit, line 9: can you change logger.error to logger.errorRequest(req, ...
this should give us more info when diagnosing errors from the logs

views/error.html
```diff
@@ -0,0 +1,23 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+  Sorry, there is a problem with the service - GOV.UK
+{% endblock %}
+
+{% block backLink %}
+  {{ govukBackLink({
+    text: "Back",
```
mwestacott, line 9: Link takes you back to the beginning / home screen rather than the previous screen so you might want to check that the text displayed for this link is correct (not like on all the other screens where the link does take you back).

SVarallo: After Talking to Oceanne as it is a system/service error and not due to Customer error it would be best to remove the back link entirely. This is also done in other services such as apply-to-protect.

PR #42, opened 2020-07-10
OBJ-69 Add sdk upload

ICoxCH:


test/services/objection.service.spec.unit.ts
```diff
@@ -35,12 +40,35 @@ describe("objections API service unit tests", () => {
...
+    line_1: "line1",
+    line_2: "line2",
+    postCode: "post code",
+  },
+  companyName: "Girls school trust",
+  companyNumber: "00006400",
+  companyStatus: "Active",
+  companyType: "limited",
+  incorporationDate: "26 June 1872",
+};
```
mwestacott, line 74: New line needed at end of file.

src/modules/sdk/objections/service.ts
```diff
@@ -48,13 +53,21 @@ export const patchObjection = async (
...
-                              fileName: string) => {
+export const addAttachment = async (companyNumber: string,
+                                    token: string,
+                                    objectionId: string,
+                                    attachment: Buffer,
+                                    fileName: string) => {
+
+  const axiosConfig: AxiosRequestConfig = getBaseAxiosRequestConfig(
+    HTTP_POST,
+    OBJECTIONS_API_ADD_ATTACHMENT_URL(companyNumber, objectionId),
```
mattch1, line 64: Is there going to be any error handling here or are we doing that a later date on another ticket?

test/modules/sdk/objections/service.spec.unit.ts
```diff
@@ -63,4 +63,16 @@ describe("objections SDK service unit tests", () => {
...
+  it("objections API is called when posting an attachment", () => {
+
+    objectionsSdk.addAttachment("companyNumber",
+      "token",
+      "objectionId",
+      new Buffer("Buffer"),
+      "fileName",
+    );
+
+    expect(mockAxiosRequest).toBeCalled();
```
markpit, line 76: could you capture the argument (axiosConfig) passed to mockAxiosRequest and check that the file info has been set correctly? maybe check the buffers and filenames match
see https://github.com/companieshouse/extensions-web/blob/master/src/test/client/apiclient.add.attachment.spec.unit.ts 
and https://jestjs.io/docs/en/mock-functions#mock-property for example on capture args

markpit: or .calledWith(stuff) might work

test/services/objection.service.spec.unit.ts
```diff
@@ -35,12 +40,35 @@ describe("objections API service unit tests", () => {
   });
 
   it("returns undefined when adding an attachment", () => {
+    mockAddAttachment.mockImplementationOnce(() => {
+      return;
+    });
+    const session = {
+      data: {
+      },
+      } as Session;
```
markpit, line 49: alignment on } as Session

test/services/objection.service.spec.unit.ts
```diff
@@ -35,12 +40,35 @@ describe("objections API service unit tests", () => {
...
     const FILE_NAME = "test_file";
-    const attachmentResult = objectionsService.addAttachment(COMPANY_NUMBER,
-        ACCESS_TOKEN,
-        NEW_OBJECTION_ID,
+    const attachmentResult = objectionsService.addAttachment(session,
         new Buffer(""),
         FILE_NAME );
-    expect(attachmentResult).toBeUndefined();
+
+    expect(attachmentResult).toEqual(Promise.resolve());
```
markpit, line 59: i'd also add expect mockAddAttachment.toBeCalled (times(1)?)

markpit: maybe check the buffer and filename are passed to mockAddAttachment would be better.

src/modules/sdk/objections/service.ts
markpit: makeApiCall will log and re-throw the error. Best to let it propagate back to the controller where it can deal with it by calling next(e)

PR #45, opened 2020-07-14
Feature/obj 115 get attachment list service

mattch1:
Api for get attachments list. 

Updated:
document upload controller.
ajax upload responder
html upload responder strategy

by replacing any[] with Attachment[] and updated the unit tests.


src/modules/sdk/objections/service.ts
```diff
@@ -14,6 +14,9 @@ const OBJECTIONS_API_PATCH_URL = (companyNumber: string, objectionId: string): s
 const OBJECTIONS_API_ADD_ATTACHMENT_URL = (companyNumber: string, objectionId: string): string =>
   OBJECTIONS_API_URL(companyNumber) + `/${objectionId}/attachments`;
 
+const OBJECTIONS_API_GET_ATTACHMENTS_LIST_URL = (companyNumber: string, objectionId: string): string =>
+    OBJECTIONS_API_URL(companyNumber) + `/${objectionId}/attachments`;
```
ICoxCH, line 18: May be better as it is identical to OBJECTIONS_API_ADD_ATTACHMENT_URL to combine them under the name OBJECTIONS_API_ATTACHMENT_URL

mattch1: Updated

PR #46, opened 2020-07-14
Feature/obj 119 file auto update bug

mattch1:
Ajax upload strategy was missing an await keyword.


mattch1:
Forget OBJ-115 has not been merged yet - so will add it to that branch.

PR #48, opened 2020-07-16
OBJ-13 remove screen display filename

mattch1:
Take the user to the "Are you sure you want to remove (attachment name)?" screen


src/controllers/remove.document.controller.ts
```diff
@@ -1,6 +1,18 @@
+import { Session } from "ch-node-session-handler";
 import { NextFunction, Request, Response } from "express";
 import { Templates } from "../model/template.paths";
+import { Attachment } from "../modules/sdk/objections";
+import * as objectionService from "../services/objection.service";
 
 export const get = async (req: Request, res: Response, next: NextFunction) => {
-  return res.render(Templates.REMOVE_DOCUMENT);
+  const attachments: Attachment[] = await objectionService.getAttachments(req.session as Session);
```
ICoxCH, line 8: Don't hate me but in the spec for the api is a {objectionId}/attachments/{attachmentId} endpoint
https://github.com/companieshouse/strike-off-objections-api/blob/master/spec/schema.json#L183
We should probably use that endpoint for this task. But the endpoint doesn't yet exist, if you change this pr against the schema, i'll implement that endpoint

mattch1: Negotiation: Could we keep it as it is for now and I can put a TODO in there stating what you ahve said here?

mattch1: Added the TODO

mattch1: Removed TODO, adde dnew single attachment service api call.

PR #53, opened 2020-07-20
OBJ-132 Implement yes journey

ICoxCH:


src/modules/sdk/objections/service.ts
```diff
@@ -106,3 +106,16 @@ export const getAttachment = async (companyNumber: string,
   const response = await makeAPICall(axiosConfig);
   return response.data as Attachment;
 };
+
+export const deleteAttachment = async (companyNumber: string,
+                                       token: string,
+                                       objectionId: string,
+                                       attachmentId: string): Promise<void> => {
```
markpit, line 113: you should be able to remove the Promise<void> return type. By marking it async it should auto wrap the return type in a promise and if left blank should return void anyway.

src/modules/sdk/objections/service.ts
```diff
@@ -106,3 +106,16 @@ export const getAttachment = async (companyNumber: string,
...
+
+  const axiosConfig: AxiosRequestConfig = getBaseAxiosRequestConfig(
+    HTTP_DELETE,
+    OBJECTIONS_API_SINGLE_ATTACHMENT_URL(companyNumber, objectionId, attachmentId),
+    token);
+  await makeAPICall(axiosConfig);
```
markpit, line 119: i think return await makeAPIcall will remove the need to promise.resolve

test/controller/document_upload/remove.document.controller.spec.unit.ts
```diff
@@ -85,10 +87,27 @@ describe("remove document url tests", () => {
...
+      .post(pageURLs.OBJECTIONS_REMOVE_DOCUMENT)
+      .send({
+        [REMOVE_DOCUMENT_FORM_FIELD]: "yes",
+        [ATTACHMENT_ID_FORM_FIELD]: ATTACHMENT_ID,
+        },
+      )
       .set("Referer", "/")
       .set("Cookie", [`${COOKIE_NAME}=123`]);
 
+    expect(mockDeleteAttachment).toBeCalledTimes(1);
```
markpit, line 110: could you also check the correct attachment id is passed to the mockDeleteAttachment

test/modules/sdk/objections/service.spec.unit.ts
```diff
@@ -98,4 +98,13 @@ describe("objections SDK service unit tests", () => {
...
     expect(mockAxiosRequest).toBeCalled();
   });
+
+  it("should call objections API deleting single attachment", () => {
+    objectionsSdk.deleteAttachment(COMPANY_NUMBER,
+      ACCESS_TOKEN,
+      OBJECTION_ID,
+      ATTACHMENT_ID);
+
+    expect(mockAxiosRequest).toBeCalled();
```
markpit, line 108: we should be testing for more here, like is the correct endpoint called with the correct values. I know this is similar to other tests like getAttachments which is also lacking, so I was going to pick this up and fix them. So leave as is and I will add tests later

test/services/objection.service.spec.unit.ts
```diff
@@ -39,15 +40,15 @@ const mockAttachment = {
     name: "test.doc",
 };
 
-mockGetAttachments.prototype.constructor.mockImplementation(async (companyNumber: string, token: string,
-                                                                   objectionId: string): Promise<Attachment[]> => {
+mockGetAttachments.mockImplementation(async (companyNumber: string, token: string,
```
markpit, line 43: if the mock returns a promise, you can use mockResolvedValue/Once (mockAttachments)
for non async functions you can use mockReturnValue.

test/services/objection.service.spec.unit.ts
```diff
@@ -132,6 +133,11 @@ describe("objections API service unit tests", () => {
         ATTACHMENT_ID);
     expect(attachment).toEqual(mockAttachment);
   });
+
+  it("should call sdk when deleting attachment", async () => {
+    await objectionsService.deleteAttachment(session, ATTACHMENT_ID);
+    expect(mockDeleteAttachment).toBeCalledTimes(1);
```
markpit, line 139: I would also check the correct attachmentId is passed as well, try toBeCalledWith or hasBeenCalledWith

PR #55, opened 2020-07-21
OBJ-127 Navigate to 'check' screen and populate with objection details

mwestacott:


src/controllers/check.your.answers.controller.ts
```diff
@@ -1,16 +1,25 @@
...
 import { NextFunction, Request, Response } from "express";
 import ObjectionCompanyProfile from "../model/objection.company.profile";
 import { Templates } from "../model/template.paths";
+import { Objection } from "../modules/sdk/objections";
+import { getObjection } from "../services/objection.service";
 import { retrieveCompanyProfileFromObjectionSession } from "../services/objection.session.service";
 import logger from "../utils/logger";
 
-export const get = async (req: Request, res: Response, next: NextFunction) => {
+export const get = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
```
markpit, line 10: you should just be able to declare the return type as void, and as it's an async function, will automatically wrap the result in a Promise.

markpit: saying that, the default function return type is void/undefined anyway

mwestacott: So do you think it needs changing then?

src/controllers/check.your.answers.controller.ts
```diff
@@ -1,16 +1,25 @@
...
   if (req.session) {
     try {
       const company: ObjectionCompanyProfile = retrieveCompanyProfileFromObjectionSession(req.session);
```
markpit, line 13: I know it's not your change but a trick you can use if you just want values from an object (like company)
const {companyName, companyNumber} = retrieveCompanyProfileFromObjectionSession(req.session);

then you can add them to res.render just as 
companyName,
companyNumber,
objection, etc..

markpit: you may need to use 'as ObjectionCompanyProfile' on teh end

mwestacott: 👍  will implement.

test/controller/check.your.answers.controller.spec.unit.ts
```diff
@@ -42,13 +45,18 @@ mockObjectionSessionMiddleware.mockImplementation((req: Request, res: Response,
...
+const mockGetObjection = getObjection as jest.Mock;
+
 describe("check company tests", () => {
 
   it("should render the page with company data from the session", async () => {
 
     mockGetObjectionSessionValue.mockReset();
     mockGetObjectionSessionValue.mockImplementation(() => dummyCompanyProfile);
 
+    mockGetObjection.mockReset();
```
markpit, line 57: you can chain these together, mock.mockReset.mockImplementation
or even better use mockResolvedValue(dummyCompanyProfile) instead of implementation

test/services/objection.service.spec.unit.ts
```diff
@@ -40,10 +41,29 @@ const mockAttachment = {
...
+    }],
+  id: "OBJ123",
+  reason: "Owed some money",
+};
+
 mockGetAttachments.mockResolvedValue(mockAttachments);
 
 mockGetAttachment.mockResolvedValue(mockAttachment);
 
+mockGetObjection.prototype.constructor.mockImplementation(async (companyNumber: string,
```
markpit, line 61: try using mockGetObjection.mockResolvedValue(mockObjection) instead.

src/controllers/check.your.answers.controller.ts
markpit: personally, i would have left it as the original version. If you want to be more explicit, : void should be fine

mwestacott: Ok, I'll revert.

test/controller/check.your.answers.controller.spec.unit.ts
mwestacott: The chaining works but using mockResolvedValue(dummyCompanyProfile) doesn't. Weird because it does for the dummyObjection call. Have left both as 'mockImplementation' for consistency.

markpit: mockGetObjection.mockReset().mockResolvedValue(dummyObjection);
works for me, make sure the param is the object and not a function

mwestacott: Yep, that worked for me too, but 'mockGetObjectionSessionValue.mockReset().mockResolvedValue(dummyCompanyProfile)' doesn't work for some reason, which is why I left them the same.

markpit: that would need a mockReturnValue/Once as it's not returning a promise.

mwestacott: Oh right. Fixed now.

src/services/objection.service.ts
```diff
@@ -103,3 +103,21 @@ const getValuesForApiCall = (session: Session) => {
...
+
+/**
+ * Gets objection details for the company and objection present in the web session.
+ *
+ * @param {Session} session the web session
+ *
+ * @returns {Objection} details of the existing objection
+ */
+export const getObjection = async (session: Session): Promise<Objection> => {
+  const companyProfileInSession: ObjectionCompanyProfile = retrieveCompanyProfileFromObjectionSession(session);
```
markpit, line 115: you should be able to replace these 4 lines with 
const { objectionId, companyNumber, token } = getValuesForApiCall(session);
after the master merge.
Also move the function up above the non exported one to keep it grouped with the others

test/modules/sdk/objections/service.spec.unit.ts
```diff
@@ -160,6 +160,19 @@ describe("objections SDK service unit tests", () => {
       HTTP_DELETE,
     );
   });
+
+  it("should call objections API getting an objection", () => {
+    objectionsSdk.getObjection(COMPANY_NUMBER,
```
markpit, line 165: i think you should check that the function returns the object as expected and call with await (make this test function async)

mwestacott: Done.

mattch1:
Check Mark and Ian are happy before merging.

PR #58, opened 2020-07-23
Add submodule updates for certified copies

rjadhav-ch:
Adds updated submodule with changes for certified copies (GCI-989)


VaughanJackson123:
Seems this one was stuck on an earlier version of `api-enumerations` than the others prior to your update?

PR #63, opened 2020-07-24
Feature/obj 17 submitted screen

mattch1:
Confirmation screen - view, routing and testing


src/model/page.urls.ts
```diff
@@ -26,4 +27,5 @@ export const OBJECTIONS_DOCUMENT_UPLOAD: string = STRIKE_OFF_OBJECTIONS + DOCUME
 export const OBJECTIONS_DOCUMENT_UPLOAD_CONTINUE: string = STRIKE_OFF_OBJECTIONS + DOCUMENT_UPLOAD_CONTINUE;
 export const OBJECTIONS_REMOVE_DOCUMENT: string = STRIKE_OFF_OBJECTIONS + REMOVE_DOCUMENT;
 export const OBJECTIONS_CHECK_YOUR_ANSWERS: string = STRIKE_OFF_OBJECTIONS + CHECK_YOUR_ANSWERS;
+export const OBJECTIONS_CONFIRMATION: string = STRIKE_OFF_OBJECTIONS +  CONFIRMATION;
```
mwestacott, line 30: Double space before 'CONFIRMATION'.

test/controller/confirmation.controller.spec.unit.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Session } from "ch-node-session-handler";
+import { NextFunction, Request, Response } from "express";
+import request from "supertest";
+import app from "../../src/app";
+import { OBJECTIONS_SESSION_NAME } from "../../src/constants";
+import authenticationMiddleware from "../../src/middleware/authentication.middleware";
+import objectionSessionMiddleware from "../../src/middleware/objection.session.middleware";
+import sessionMiddleware from "../../src/middleware/session.middleware";
+import { OBJECTIONS_CONFIRMATION } from "../../src/model/page.urls";
+import {
```
mwestacott, line 16: Indentation throughout this file looks wrong - 4 spaces rather than 2.

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,22 @@
...
+import { NextFunction, Request, Response } from "express";
+import { SESSION_OBJECTION_ID } from "../constants";
+import { Templates } from "../model/template.paths";
+import { retrieveFromObjectionSession, retrieveUserProfileFromSession } from "../services/objection.session.service";
+
+export const get = (req: Request, res: Response, next: NextFunction) => {
+    const session: Session | undefined  = req.session;
+    if (session) {
+        const userProfile = retrieveUserProfileFromSession(session);
+        const objId: string = retrieveFromObjectionSession(session, SESSION_OBJECTION_ID);
```
mwestacott, line 12: You could call this variable 'objectionId' and then just reference it in the render method below, rather than do 'objectionId: objId'. Just a short-hand way of achieving the same thing. Take a look at the check your answers controller which does something similar.

test/controller/confirmation.controller.spec.unit.ts
mattch1: 2 spaces

src/controllers/check.your.answers.controller.ts
```diff
@@ -24,6 +23,9 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
       return next(e);
     }
   }
-
   return next(new Error("No Session present"));
 };
+
+export const post = (req: Request, res: Response) => {
+  res.redirect(OBJECTIONS_CONFIRMATION);
```
markpit, line 30: add a return to this to prevent anything executing afterwards

src/controllers/confirmation.controller.ts
mattch1: changed this to use variables only

src/model/page.urls.ts
mattch1: removed extra space

src/controllers/check.your.answers.controller.ts
mattch1: return added

src/controllers/confirmation.controller.ts
```diff
@@ -0,0 +1,23 @@
...
+import { NextFunction, Request, Response } from "express";
+import { SESSION_OBJECTION_ID } from "../constants";
+import { Templates } from "../model/template.paths";
+import { retrieveFromObjectionSession, retrieveUserProfileFromSession } from "../services/objection.session.service";
+
+export const get = (req: Request, res: Response, next: NextFunction) => {
+    const session: Session | undefined  = req.session;
+    if (session) {
+        const userProfile = retrieveUserProfileFromSession(session);
+        const email: string = userProfile[UserProfileKeys.Email] as string;
```
markpit, line 12: rather than getting the user profile to get the email from it, why not just have a getEmail function on session service

mattch1: I did consider this but it just seemed better to get the whole user profile back as it has many fields and makes it more reusable.

markpit: the downside is that we have session knowledge leaking out into the controllers, they will have to have knowledge of the construction of the userProfile. Anyway that's my 2p.

test/services/objection.session.service.spec.unit.ts
```diff
@@ -1,10 +1,37 @@
...
+      },
+    };
+    const userProfile: IUserProfile = retrieveUserProfileFromSession(session);
+    expect(userProfile).not.toBeUndefined();
+    expect(userProfile).toEqual({email: "demo@ch.gov.uk", forename: "Joe", surname: "Bloggs"});
+  });
+
+  it("should throw error when user profile is absent", () => {
+    const session: Session = new Session();
+    expect(() => {
```
markpit, line 30: this throws when there is no signIninfo, to test if it throws when there is no user profile you would need a session with signInInfo but no user profile.

test/controller/confirmation.controller.spec.unit.ts
mwestacott: Most of the indentation below is still at 4 spaces.

src/controllers/confirmation.controller.ts
mattch1: Alright I've updated this.

test/controller/confirmation.controller.spec.unit.ts
mattch1: Mdae all four space two spaces  there wa another file like this as well.

test/services/objection.session.service.spec.unit.ts
mattch1: Updated tests.

PR #66, opened 2020-07-28
OBJ-158 Submit objection after confirming details

mwestacott:


src/modules/sdk/objections/types.ts
```diff
@@ -14,8 +14,8 @@ export interface ApiError {
  */
 export enum ObjectionStatus {
   OPEN,
-  SUBMITTED,
   PROCESSED,
+  SUBMITTED,
```
mwestacott, line 18: So this caused me some grief and confusion with its original ordering - I was setting ‘SUBMITTED’ in the web but the API was receiving and setting ‘PROCESSED’. Turns out that Javascript doesn’t really support enums and the typescript was generating code that meant what’s really set is an index value and not a string enum value. So… I set ‘SUBMITTED’ in the web which got sent to the API as ‘1’ (2nd item in web enum) which in turn got decoded in the API to ‘PROCESSED’ (as this is the 2nd item in the API ObjectionStatus enum). Upshot is that Javascript/Java enums duplicated between web and API need to have the same ordering. Hope this makes sense and you agree with this synopsis?

(BTW went back and checked the PTF enums and it seems we did also align the ordering there too)

test/controller/check.your.answers.controller.spec.unit.ts
```diff
@@ -68,13 +72,16 @@ describe("check company tests", () => {
...
     const res = await request(app)
         .post(OBJECTIONS_CHECK_YOUR_ANSWERS)
         .set("referer", "/")
         .set("Cookie", [`${COOKIE_NAME}=123`]);
+
+    expect(mockSubmitObjection).toBeCalledWith(dummySession);
+
     expect(res.status).toEqual(302);
-    expect(res.text).toContain(OBJECTIONS_CONFIRMATION);
+    expect(res.header.location).toEqual(OBJECTIONS_CONFIRMATION);
```
mwestacott, line 84: We've done this elsewhere. I think it's a more rigorous check than just examining the page text.

src/modules/sdk/objections/types.ts
markpit: ah, what we need are sting enums then:
enum Direction {
    Up = "UP",
    Down = "DOWN",
    Left = "LEFT",
    Right = "RIGHT",
}
https://www.typescriptlang.org/docs/handbook/enums.html

mwestacott: Ah right. Will change and try this approach.

mwestacott: Yep, that works. Updated.

PR #67, opened 2020-08-06
OBJ-181 htmlencode user input before rendering to screen

ICoxCH:


test/routes/routes.spec.unit.ts
```diff
@@ -126,3 +132,14 @@ const dummyCompanyProfile: ObjectionCompanyProfile = {
...
+
+const dummyObjection: Objection = new Objection(
+  "Owed some money",
+  [
+    {
+      name: "attachment.jpg",
+    },
+    {
+      name: "document.pdf",
+    }],
```
SVarallo, line 144: Formatting on the ]

package.json
```diff
@@ -25,6 +25,7 @@
     "express": "~4.16.1",
     "form-data": "^2.3.3",
     "govuk-frontend": "^3.6.0",
+    "he": "^1.2.0",
```
mwestacott, line 28: What a great name for a library!

src/modules/sdk/objections/types.ts
```diff
@@ -38,12 +40,44 @@ export interface Attachment {
...
+  get reason(): string {
+    return this._reason;
+  }
+
+  get attachments(): Array<{ name: string }> {
+    return this._attachments;
+  }
+
+  public getHtmlEncoded(): Objection {
+    const htmlReason: string = he.encode(this._reason);
```
mwestacott, line 66: Would it be an option - and is it worth it - to HTML encode the data on object construction and store the encoded fields in the object rather than encode them every time they're requested.

Guessing this will come down to how often the 'getHtmlEncoded()' method is called (and I suspect not often) and how expensive it is.

Just though I should ask the question...

ICoxCH: OWASP seem to favour only encoding on output, and encoding for where it is going to be used.
If we only had the htmlEncoded version stored and needed to pass the name into js, it'd be a lot of effort to unhtml encode it, only to re URIencode it.

mwestacott: I was thinking about storing both the encoded and unencoded output actually. So there'd be more fields in the object class.

markpit: I like that idea, what if we had a setter that sets the normal field data but also encodes and populates an encoded version as another field. So 1 setter populates 2 fields. Then have 2 getters, a normal one, and getEncodedX version. That way we'll only have to run the encoder once per field.

markpit: or, switch them around so that eg. getReason would return htmlEncoded version, and getRawReason would return the 'raw' un-encoded version (but still do the encoding in the constructor)

ICoxCH:
Turns out I was overcomplicating everything, was more of a bug where we were passing variables to the summary list as html, so the browser was parsing them as html.
I've set the fields to text and the Nunjuks autoescaping will handle the rest for us

mwestacott:
> Turns out I was overcomplicating everything, was more of a bug where we were passing variables to the summary list as html, so the browser was parsing them as html.
> I've set the fields to text and the Nunjuks autoescaping will handle the rest for us

Much simpler! 😊

PR #68, opened 2020-08-06
OBJ-101 error handling for invalid file types

mattch1:
Error message code to handle 415 status error from the api.

### DIFF
--- a/src/controllers/document_upload/document.upload.controller.ts
+++ b/src/controllers/document_upload/document.upload.controller.ts
@@ -63,7 +63,7 @@ export const postFile = async (req: Request, res: Response, next: NextFunction) 
   const uploadCallbacks: UploadFileCallbacks = {
     fileSizeLimitExceededCallback: getFileSizeLimitExceededCallback(req, res, uploadResponderStrategy, attachments),
     noFileDataReceivedCallback: getNoFileDataReceivedCallback(req, res, uploadResponderStrategy, attachments),
-    uploadFinishedCallback: getUploadFinishedCallback(req, res, next, uploadResponderStrategy),
+    uploadFinishedCallback: getUploadFinishedCallback(req, res, next, uploadResponderStrategy, attachments),
   };
   const maxFileSizeBytes: number = parseInt(MAX_FILE_SIZE_BYTES, 10);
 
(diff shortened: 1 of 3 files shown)


src/controllers/document_upload/document.upload.controller.ts
```diff
@@ -157,10 +159,10 @@ const getUploadFinishedCallback = (req: Request,
 
       // TODO OBJ-101 - file type error - uncomment below should do it
```
mwestacott, line 160: Guessing that TODO comment should now be deleted if this is the full and final implementation of OBJ-101.

src/controllers/document_upload/document.upload.controller.ts
```diff
@@ -157,10 +159,10 @@ const getUploadFinishedCallback = (req: Request,
...
       // // render errors in the view
-      // if (e.status === 415) {
-      //   return await
-      //     displayError(req, res, UploadErrorMessages.INVALID_MIME_TYPES, uploadResponderStrategy, attachments);
-      // }
+      if (e.status === 415) {
```
mwestacott, line 162: 415 is a magic number! Would be good to find out what this error code actually means, define a well-named constant for it and use that.

mattch1: Not sure what you mean by magic number - but will extract a constant if there are no references to this.

mwestacott: https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

And thanks.

test/controller/document_upload/document.upload.controller.spec.unit.ts
```diff
@@ -374,7 +374,7 @@ describe ("document.document_upload.controller tests", () => {
       throw {
         data: ["error"],
         message: "error",
-        status: 415,
+        status: HttpStatusCodes.UNSUPPORTED_MEDIA_TYPE,
```
markpit, line 377: could you change this back to 415, the real error code will be a number , just want to make sure that when we do the comparison in the code, it works when comparing the number to the enum.

markpit: just in the test i mean

src/controllers/document_upload/document.upload.controller.ts
mattch1: Updated this.

src/controllers/document_upload/document.upload.controller.ts
mattch1: Removed the TODO

test/controller/document_upload/document.upload.controller.spec.unit.ts
mattch1: Ok chnaged back to "415".

PR #69, opened 2020-08-10
OBJ-205 Adding stub for api download call

markpit:
Also updated the objections service to call the sdk for download.


src/modules/sdk/objections/service.ts
```diff
@@ -121,6 +122,12 @@ export const deleteAttachment = async (companyNumber: string,
   return;
 };
 
+export const downloadAttachment = async (downloadUri: string,
+                                         httpResponse: Response,
+                                         token: string) => {
+  return Promise.resolve();
```
mwestacott, line 128: I'd suggest adding a TODO comment that references the sub-task which describes replacing this stub with the actual implementation.

markpit: updated

PR #70, opened 2020-08-11
OBJ-193 adding download landing page and controller

SVarallo:
Jira ticket at https://companieshouse.atlassian.net/browse/OBJ-193

### DIFF
--- a/src/controllers/download.attachment.landing.controller.ts
+++ b/src/controllers/download.attachment.landing.controller.ts
@@ -1,4 +1,5 @@
 import { NextFunction, Request, Response } from "express";
+import { DOWNLOAD_PREFIX } from "../model/page.urls";
 import { Templates } from "../model/template.paths";
 import logger from "../utils/logger";
 
@@ -9,7 +10,8 @@ import logger from "../utils/logger";
  * @param next
  */
 export const get = (req: Request, res: Response, next: NextFunction) => {
-  const url: string = req.originalUrl.replace("download/", "");
+  // replaces first match only
+  const url: string = req.originalUrl.replace(DOWNLOAD_PREFIX, "");
   logger.debug("Download landing page with download url = " + url);
   return res.render(Templates.DOWNLOAD_ATTACHMENT_LANDING_PAGE, {downloadUrl: url});
 };
--- a/src/controllers/download.attachment.landing.controller.ts
+++ b/src/controllers/download.attachment.landing.controller.ts
@@ -1,0 +1,15 @@
+import { NextFunction, Request, Response } from "express";
+import { Templates } from "../model/template.paths";
+import logger from "../utils/logger";
+
+/**
+ * GET controller for download attachment landing screen
+ * @param req
+ * @param res
+ * @param next
+ */
+export const get = (req: Request, res: Response, next: NextFunction) => {
+  const url: string = req.originalUrl.replace("download/", "");
+  logger.debug("Download landing page with download url = " + url);
+  return res.render(Templates.DOWNLOAD_ATTACHMENT_LANDING_PAGE, {downloadUrl: url});
+};
--- a/src/model/page.urls.ts
+++ b/src/model/page.urls.ts
@@ -13,6 +13,10 @@ export const ENTER_INFORMATION: string = SEPARATOR + Templates.ENTER_INFORMATION
 export const DOCUMENT_UPLOAD: string = SEPARATOR + Templates.DOCUMENT_UPLOAD;
 export const DOCUMENT_UPLOAD_CONTINUE: string = SEPARATOR + "document-upload-continue"; // not a template
 export const REMOVE_DOCUMENT: string = SEPARATOR + Templates.REMOVE_DOCUMENT + "/:attachmentId";
+export const DOWNLOAD_ATTACHMENT: string =
+  "/company/:companyId/strike-off-objections/:requestId/attachments/*/download";
+export const DOWNLOAD_PREFIX: string = SEPARATOR + Templates.DOWNLOAD_ATTACHMENT_LANDING_PAGE;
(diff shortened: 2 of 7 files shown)


src/model/page.urls.ts
```diff
@@ -13,6 +13,10 @@ export const ENTER_INFORMATION: string = SEPARATOR + Templates.ENTER_INFORMATION
 export const DOCUMENT_UPLOAD: string = SEPARATOR + Templates.DOCUMENT_UPLOAD;
 export const DOCUMENT_UPLOAD_CONTINUE: string = SEPARATOR + "document-upload-continue"; // not a template
 export const REMOVE_DOCUMENT: string = SEPARATOR + Templates.REMOVE_DOCUMENT + "/:attachmentId";
+export const DOWNLOAD_ATTACHMENT: string =
+  "/company/:companyId/strike-off-objections/:requestId/attachments/*/download";
```
ICoxCH, line 17: Could be worth doing :attachmentId instead of * so it is clear what the number is, and we can access the field if need be

test/controller/download.attachment.landing.controller.spec.unit.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const mockSessionMiddleware = sessionMiddleware as jest.Mock;
+mockSessionMiddleware.mockImplementation((req: Request, res: Response, next: NextFunction) => {
+  req.session = {
+    data: {},
+  } as Session;
+  return next();
+});
+
+const mockObjectionSessionMiddleware = objectionSessionMiddleware as jest.Mock;
```
ICoxCH, line 28: As this will only be accessed by internal staff (I believe) we should probably exclude this middleware from running against this endpoint as an ObjectionSession isn't really useful for them

src/model/page.urls.ts
```diff
@@ -13,6 +13,10 @@ export const ENTER_INFORMATION: string = SEPARATOR + Templates.ENTER_INFORMATION
...
+export const DOWNLOAD_PREFIX: string = SEPARATOR + Templates.DOWNLOAD_ATTACHMENT_LANDING_PAGE;
```
markpit, line 18: the prefix and the template name are 2 separate things even though they might share the same value. I would make this a string literal  SEPARATOR + "download"

test/controller/download.attachment.landing.controller.spec.unit.ts
```diff
@@ -0,0 +1,69 @@
...
+mockObjectionSessionMiddleware.mockImplementation((req: Request, res: Response, next: NextFunction) => {
+  if (req.session) {
+    req.session.data[OBJECTIONS_SESSION_NAME] = {};
+    return next();
+  }
+
+  return next(new Error("No session on request"));
+});
+
+const DOWNLOAD_LANDING_PAGE_URL: string = "/strike-off-objections/download/company/1234/strike-off-objections/5678/attachments/8888/download";
```
markpit, line 38: the real url will start with the download prefix, i would add that to the url

SVarallo: It does have the download prefix, just after the service name.

src/controllers/download.attachment.landing.controller.ts
```diff
@@ -0,0 +1,15 @@
...
+import logger from "../utils/logger";
+
+/**
+ * GET controller for download attachment landing screen
+ * @param req
+ * @param res
+ * @param next
+ */
+export const get = (req: Request, res: Response, next: NextFunction) => {
+  const url: string = req.originalUrl.replace("download/", "");
```
markpit, line 12: If i remember , i think this is stripping off the download prefix, in which case you could reference the value from pageUrls PREFIX + SEPARATOR?

test/controller/download.attachment.landing.controller.spec.unit.ts
markpit: oh yes , i do apologise

test/controller/download.attachment.landing.controller.spec.unit.ts
markpit: we will need the user's access token from the session (not chSession though? not sure) to call the api. I think this is how we get the eric user info to authorise the download on the api side.

src/controllers/download.attachment.landing.controller.ts
SVarallo: Replaced with the DOWNLOAD_PREFIX from pageUrls, .replace only removes the first instance of /download so it won't trim off the end as well.

src/model/page.urls.ts
SVarallo: Replaced the template name reference with a string literal.

src/model/page.urls.ts
SVarallo: Replaced

test/controller/download.attachment.landing.controller.spec.unit.ts
markpit: looking at the code, we have 2 middleware functions for the session, the objection.session one creates the 'objection session' add on data and won't affect the access token. So you could do as Ian suggested and add an 'if' into the objection.session middleware to return next if it is a download call. You could do it in this PR or a separate one, up to you.

markpit: if you do manage to bypass the objection.session middleware, you could actually keep that mock and test that the mock isn't called for a download.

ICoxCH: I think if you change app.ts
`app.use(`${pageURLs.STRIKE_OFF_OBJECTIONS}/*`, objectionSessionMiddleware);` to 
`app.use(`${pageURLs.STRIKE_OFF_OBJECTIONS}/*(?<!download)$`, objectionSessionMiddleware);`
then objectionSessionMiddleware will not run on endpoints ending in download
regexr.com/5a2k8  (If node supports negative lookbehind that is, a quick google suggests it might)

PR #71, opened 2020-08-12
OBJ-195 Generic file error template

mattch1:
Modelled on the extensions example - a generic  file error template ready to take a forbidden error message.

### DIFF
--- a/views/file-error.html
+++ b/views/file-error.html
@@ -1,14 +1,14 @@
 {% extends "layout.html" %}
 
 {% block pageTitle %}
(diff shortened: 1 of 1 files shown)


views/file-error.html
```diff
@@ -0,0 +1,14 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+{{heading}} - GOV.UK
```
markpit, line 4: indentation

mattch1: Updated this.

PR #72, opened 2020-08-16
OBJ-194, OBJ-215 Implement sdk download

markpit:
Adding the real download implementation to replace the stub
Adding download controller and service implementation
Files downloaded will be prefixed with a value from config to help users identify what files are associated with this app.


src/controllers/document.download.controller.ts
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Show the correct error page to the user
+ * @param {number} httpStatus
+ * @param {Response} res
+ */
+const showErrorPage = (httpStatus: number, res: Response) => {
+  const { UNAUTHORIZED, FORBIDDEN, NOT_FOUND } = HttpStatusCodes;
+  switch (httpStatus) {
+    case UNAUTHORIZED: {
```
ICoxCH, line 77: I've just had the thought, from a security perspective if they are unauthorized or forbidden, we should probably act as if the file wasnt found.
This way a malicious user can use the "Not authorized" response to know that a file of given id exists.

markpit: I think Sarah has already discussed with the business what messages will be shown - she's created a story for the forbidden one - https://companieshouse.atlassian.net/browse/OBJ-190. I can let her know.

ICoxCH: Mayb be worth double checking with Steve or Kev as user requirements may clash with security recommendations here

markpit: updated, does this mean we'll have to change the API to return 404 as well

ICoxCH: I suppose we should, though as the api is internal that could be relatively low priority?

mwestacott: 👍 should create a tech story to do this, so it's not forgotten. I'll do this if you like.

src/model/page.urls.ts
```diff
@@ -13,10 +13,10 @@ export const ENTER_INFORMATION: string = SEPARATOR + Templates.ENTER_INFORMATION
 export const DOCUMENT_UPLOAD: string = SEPARATOR + Templates.DOCUMENT_UPLOAD;
 export const DOCUMENT_UPLOAD_CONTINUE: string = SEPARATOR + "document-upload-continue"; // not a template
 export const REMOVE_DOCUMENT: string = SEPARATOR + Templates.REMOVE_DOCUMENT + "/:attachmentId";
-export const DOWNLOAD_ATTACHMENT: string =
+export const DOCUMENT_DOWNLOAD: string =
```
mwestacott, line 16: I can see why you changed this to standardise on the terminology but we do still refer to attachments in a number of places, so still have a mixture of terms going on. Probably in the API too.

markpit: i know, just wanted the controllers aligned, kept it as attachment in function names as that's what it is called in the API. Without changing the api , i tried to get it as close as I could.

PR #74, opened 2020-08-17
Feature/obj 29 migrate lint to eslint

mattch1:
Removed ts lint config.
Added es lint config.
Auto corrected a number of files.

There were a number of warnings and errors when this was run, the vast majority were to do with indentation when the 2 space rule was set. Used the auto correct tool in Goland to fix them - this is why there are a number of changes in the code alongside the config changes. Also the string types for string literals were removed.

Whitespace rules some enabled by default, others can be fixed with the ide tool for eslint.

Switched four rules off:

**No explicit any:** any is used in the codebase – having it explicit makes the code more readable. 9 problems highlighted.

**No unused vars**: Too indiscriminate, some variables are required in the code even though they're being marked as never used. For example variables beginning with under score in the ajax upload responder strategy. 23 problems highlighted.

**Explicit module boundary types**: Decisions needed: not clear whether the return Keyword has been left out for a reason. For example, in objection service there are a number of functions without return types. 31 problems highlighted.

**Ban types**: Decisions needed:  object arrays are used in the ajax upload responder strategy, apparently there are problems when it comes to accessing them with keys (https://github.com/microsoft/TypeScript/issues/21732). Some decision needs to be taken about whether we continue to use this. 4 problems highlighted.


ICoxCH:
I question whether turning those 4 rules off is the correct decision rather than investigating and addressing these issues.
The explicit module boundary types, with missing returns sounds like an issue that should be addressed.

I think the unused vars rule should be turned on, and if absolutely needed, we can exclude those lines in the ajax responder strategy (though if eslint is saying they are unused this could use further investigation)

The No explicit any rule, looking at the companies house standard says to use JS standard, and if that says "No explicit any" then we should probably stick to that.

Ban types could use some investigation into whether we can refactor that method to be eslint compliant.

I suppose the real question is should refactoring our code base to conform to this eslint standard come under this ticket or another one?
Though I believe we should not turn any of these rules off, happy for more thoughts on this however

mattch1:
I did look into addressing the issues and continue to do so, but have not yet found any way to refine these rules. In other cases I think a collective decision should be taken particularly with the ban-types rule. I did not just simply opt to turn off the rules - I have just put a stopper on the errors for now while I investgate further and also refer this just to get some other eyes on how we refine this further.

src/controllers/document_upload/ajax.upload.responder.strategy.ts
```diff
@@ -62,8 +62,8 @@ export class AjaxUploadResponderStrategy implements UploadResponderStrategy {
    * @param {Attachment[]} attachments list of uploaded attachments
    */
   public handleGovUKError = async (res: Response,
-                                   errorData: GovUkErrorData,
-                                   attachments: Attachment[]) => {
+    errorData: GovUkErrorData,
```
markpit, line 65: You need to migrate the tslint rules that we applied over to eslint, there was a rule for param alignment.

ICoxCH:
Is there a reason for removing the package-lock?

src/controllers/document_upload/ajax.upload.responder.strategy.ts
mattch1: Migrated the rules - names and definitons very different.

mattch1:
Removed package lock temporarily as it had been committed unintentionally from this branch. Then reintroduced the package lock we have on master.

src/controllers/document.download.controller.ts
```diff
@@ -74,15 +74,15 @@ const prefixFilename = (contentDisposition: string ): string => {
...
-    case NOT_FOUND: {
-      const { HEADING_NOT_FOUND, MESSAGE_NOT_FOUND } = DownloadErrorMessages;
-      return renderFileError(res, HEADING_NOT_FOUND, MESSAGE_NOT_FOUND);
-    }
-    default: {
-      return res.render(Templates.ERROR);
-    }
+  case UNAUTHORIZED:
+  case FORBIDDEN:
+  case NOT_FOUND: {
```
markpit, line 79: This indentation doesn't look right to me, especially with the 2 braces vertically aligned below. does it complain if we have normal indentation like this 
`switch (httpStatus) {
    case UNAUTHORIZED:`

mattch1: Needed to specify for switch statements.

markpit:
> Removed package lock temporarily as it had been committed unintentionally from this branch. Then reintroduced the package lock we have on master.

The package lock is expected to change as we are removing tslint and adding eslint. Make sure to run 'make' again and that should regenerate a new one and commit that one.

src/controllers/document_upload/document.upload.controller.ts
```diff
@@ -110,7 +110,7 @@ const getFileSizeLimitExceededCallback = (req: Request,
   return async (filename: string, maxInBytes: number) => {
     const maxInMB: number = getMaxFileSizeInMB(maxInBytes);
     logger.debug("File limit " + maxInMB + "MB reached for file " + filename);
-    const errorMsg: string = `${UploadErrorMessages.FILE_TOO_LARGE} ${maxInMB} MB`;
+    const errorMsg = `${UploadErrorMessages.FILE_TOO_LARGE} ${maxInMB} MB`;
```
mwestacott, line 113: Removing the type information here seems to be a backwards step. Is there a rule that says this is a bad thing to do?

src/controllers/document_upload/document.upload.controller.ts
```diff
@@ -159,7 +159,7 @@ const getUploadFinishedCallback = (req: Request,
 
       if (e.status === HttpStatusCodes.UNSUPPORTED_MEDIA_TYPE) {
         return await
-          displayError(res, UploadErrorMessages.INVALID_MIME_TYPES, uploadResponderStrategy, attachments);
+        displayError(res, UploadErrorMessages.INVALID_MIME_TYPES, uploadResponderStrategy, attachments);
```
mwestacott, line 162: The indentation looks wrong here - it's a line continuation, but now that the indentation has gone, it starts to look like two separate lines.

src/controllers/document_upload/document.upload.controller.ts
mattch1: Not really in previous PRs - it has been recommended that we removed types in this situation.

markpit: It is ok in this situation, if you are assigning a string literal to a const/let/var as it's obvious what it is. if we were assigning the result of a function then we would need to specify :string to make it clear. There's some stuff on google about it, best practices perhaps.

mattch1: The rule for this seems to be true or false with no additional paramters to refine it, so I'll have to just switch it off.

markpit: rule looks ok to me, we don't need to declare type when assigning a string literal.

src/controllers/document_upload/document.upload.controller.ts
mattch1: Not sure how to refine this, there does not seem to be any parameter to acknowledge line continuation.

mattch1: Bit stuck here not sure how to refine the indentation rule to get it to acknowledge continued expressions.

src/controllers/document_upload/document.upload.controller.ts
mwestacott: Yep, agreed we should leave the rule.

src/controllers/document_upload/document.upload.controller.ts
markpit: if we just return the indentation to as it was on this line, will eslint be ok with it?  or just fit it all on 1 line, looks like it might fit.

mwestacott: Just googled the eslint indentation rules and also can't find a switch to control this. Perhaps try as @markpit suggests above, if not, I guess we live with this.

mattch1: Tried this, but the rule complains so we'll have to keep it as it is for now.

markpit (approved):

mwestacott (approved):

PR #75, opened 2020-08-20
OBJ-218 Bug fix to forward to original page for downloads if logged out

markpit:


src/controllers/document_upload/document.upload.controller.ts
```diff
@@ -138,7 +138,7 @@ const getNoFileDataReceivedCallback = (req: Request,
  * @param {Response} res http response
  * @param {NextFunction} next the next function in the middleware chain
  * @param {UploadResponderStrategy} uploadResponderStrategy the strategy for responding to requests
- * @param {attachments} required to display invalid mime types error
+ * @param {attachments} attachments list of attachments needed to re-draw the screen
```
markpit, line 141: ide highlighted param name missing from javascript doc so updated description as well

PR #76, opened 2020-08-21
Feature/obj 29 fix eslint upload

markpit:
eslint highlighted some async await issues in the uploader.
I've made the strategy interface functions return promises as the eslint issues arose due to the ajax implementation using async / promises and the html version didn't.
This change forces all functions implemented in the strategy interface to return promises.

Also spotted a couple of errors in the check-your-answers page (back link and extra quotation)

### DIFF
--- a/src/controllers/document_upload/ajax.upload.responder.strategy.ts
+++ b/src/controllers/document_upload/ajax.upload.responder.strategy.ts
@@ -37,7 +37,7 @@ export class AjaxUploadResponderStrategy implements UploadResponderStrategy {
         .then((html: string) =>  this.addReplacementDiv(replacementDivs, html, CHOOSE_FILE_DIV));
       logger.trace("Rendered fragment " + Templates.DOCUMENT_UPLOAD_FILE_PICKER);
 
-      res.send({divs: replacementDivs});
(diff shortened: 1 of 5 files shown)


ICoxCH (approved):

src/controllers/document_upload/ajax.upload.responder.strategy.ts
```diff
@@ -50,9 +50,10 @@ export class AjaxUploadResponderStrategy implements UploadResponderStrategy {
    * @param {Error} e the error
    * @param {NextFunction} _next the next middleware function
    */
-  public handleGenericError = (res: Response, e: Error, _next?: NextFunction) => {
+  public handleGenericError = (res: Response, e: Error, _next?: NextFunction): Promise<void> => {
     logger.error(ErrorMessages.ERROR_500 + ": " + e);
     res.status(500).send({ redirect: pageURLs.OBJECTIONS_ERROR });
+    return Promise.resolve();
```
mattch1, line 56: Why do we need these? Is this something es lint has complained about or are there problems that have resulted from the absence of this?

mattch1 (approved):
Deleted comment.

PR #77, opened 2020-08-24
Create notice expired error screen and controller

SVarallo:


src/controllers/notice.expired.controller.ts
```diff
@@ -0,0 +1,17 @@
...
+import { NextFunction, Request, Response } from "express";
+import { Templates } from "../model/template.paths";
+import {
+  retrieveCompanyProfileFromObjectionSession,
+} from "../services/objection.session.service";
+
+export const get = async (req: Request, res: Response, next: NextFunction) => {
+  const session: Session = req.session as Session;
+
+  const companyName: string = retrieveCompanyProfileFromObjectionSession(session).companyName;
```
markpit, line 11: you can shorten this line by using 
const { companyName } = retrieveCompanyProfileFromObjectionSession(session);
that says to extract companyName from retrieveCompanyProfileFromObjectionSession(session);
not sure if you'd need to add ": string" somewhere.

views/notice-expired.html
```diff
@@ -0,0 +1,31 @@
+{% extends "layout.html" %}
+
+{% block pageTitle %}
+Notice expired
```
markpit, line 4: indent Notice

src/controllers/notice.expired.controller.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const get = async (req: Request, res: Response, next: NextFunction) => {
+  const session: Session = req.session as Session;
+
+  const { companyName } = retrieveCompanyProfileFromObjectionSession(session)
+
+  return res.render(Templates.NOTICE_EXPIRED, {
+    companyName,
+    templateName: Templates.NOTICE_EXPIRED,
+  });
```
markpit, line 16: I think this controller will need it's own unit test, so we can mock retrieveCompanyProfileFromObjectionSession. to return a number and we can test that the number appears on the page.

markpit (approved):

mwestacott (approved):

PR #82, opened 2020-09-03
OBJ-251 Throw an ApiError in sdk service if unauthorised

ICoxCH:


src/modules/sdk/objections/service.ts
```diff
@@ -154,13 +155,22 @@ export const downloadAttachment = async (downloadApiUrl: string,
       Authorization: "Bearer " + token,
     },
     proxy: false,
+    validateStatus: (status) => {
+      return ((status >= 200 && status < 300) || status === 401);
```
markpit, line 159: A 401 should throw an ApiError anyway from makeApiCall

default is 
// `validateStatus` defines whether to resolve or reject the promise for a given
  // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
  // or `undefined`), the promise will be resolved; otherwise, the promise will be
  // rejected.
  validateStatus: function (status) {
    return status >= 200 && status < 300; // default
  },

plus make sure the other codes get thrown as well:
  UNAUTHORIZED = 401,
  FORBIDDEN = 403,
  NOT_FOUND = 404,
  UNSUPPORTED_MEDIA_TYPE = 415,

ICoxCH: It does throw an api error for 401 in the makeApiCall but the api error was never making it to the document download controller for it to interperet and render the correct page and I couldn't figure out where it was going

ICoxCH: hold on looking at the original error again, i may have a better solution

src/controllers/document.download.controller.ts
```diff
@@ -35,7 +36,7 @@ export const get = async (req: Request, res: Response) => {
     return download.data.pipe(res);
 
   } catch (e) {
-    logger.errorRequest(req, `download.attachment.controller - ${JSON.stringify(e)}`);
+    logger.errorRequest(req, `download.attachment.controller - ${inspect(e)}`);
```
ICoxCH, line 39: 'e' contained circular entries and Stringify cannot handle that so was falling over.
Inspect replaces any circular entries with [circular]

markpit (approved):
Looks good but don't merge until the demo is done!

src/controllers/document.download.controller.ts
mwestacott: A one line comment in the code (immediately above) explaining this would be useful I think.

mwestacott (approved):

PR #83, opened 2020-09-03
OBJ-49 creating objecting entity name screen

SVarallo:
Story at https://companieshouse.atlassian.net/browse/OBJ-49

Adding the empty post method and the error variables in the HTML in preparation for the validation that will be done in OBJ-50.

Will need to let testers know before this is merged as it will break all of their tests.


ICoxCH (approved):

mwestacott (approved):
Looks fine. As you say, need to coordinate this merge with tester changes or our pipeline will most likely break.

PR #88, opened 2020-09-14
OBJ-50 adding validation to objecting entity details screen

SVarallo:
Jira ticket at https://companieshouse.atlassian.net/browse/OBJ-50

### DIFF
--- a/src/controllers/objecting.entity.name.controller.ts
+++ b/src/controllers/objecting.entity.name.controller.ts
@@ -16,6 +16,12 @@ const validators = [
   check(DIVULGE_INFO_FIELD).not().isEmpty().withMessage(ErrorMessages.SELECT_TO_DIVULGE),
 ];
 
+/**
+ * ProcessForm validates input and processes form
+ * @param req
+ * @param res
+ * @param next
+ */
 export const processForm = [...validators, async (req: Request, res: Response, next: NextFunction): Promise<void> => {
   const errors = validationResult(req);
   const errorListData: GovUkErrorData[] = [];
--- a/src/controllers/objecting.entity.name.controller.ts
+++ b/src/controllers/objecting.entity.name.controller.ts
@@ -3,6 +3,9 @@ import { check, ValidationError, validationResult } from "express-validator/chec
 import { ErrorMessages } from "../model/error.messages";
 import { createGovUkErrorData, GovUkErrorData } from "../model/govuk.error.data";
 import { OBJECTIONS_COMPANY_NUMBER } from "../model/page.urls";
+import { addObjectionCreateToObjectionSession } from "../services/objection.session.service";
+import { ObjectionCreate } from "../modules/sdk/objections";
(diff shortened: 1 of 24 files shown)


src/controllers/objecting.entity.name.controller.ts
```diff
@@ -1,27 +1,66 @@
...
+import { check, ValidationError, validationResult } from "express-validator";
+import { ErrorMessages } from "../model/error.messages";
+import { createGovUkErrorData, GovUkErrorData } from "../model/govuk.error.data";
 import { OBJECTIONS_COMPANY_NUMBER } from "../model/page.urls";
 import { addObjectionCreateToObjectionSession } from "../services/objection.session.service";
 import { ObjectionCreate } from "../modules/sdk/objections";
 import { Session } from "ch-node-session-handler";
+import { Templates } from "../model/template.paths";
+
+const FULL_NAME_FIELD: string = "fullName";
```
markpit, line 11: when declaring string literal constants you can leave off the : string 
ie. x = "some string", you don't need : string 
x: string = getMessage() you would need it.

markpit: it's ok as it is, just a best practice thing

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -1,27 +1,66 @@
...
  * @param req
  * @param res
  * @param next
  */
-export const post = (req: Request, res: Response, next: NextFunction) => {
-  const shareIdentityString: string = req.body.shareIdentity;
-  const fullNameValue: string = req.body.fullName;
-  const shareIdentityValue: boolean = shareIdentityString === "yes";
-  const objectionCreate: ObjectionCreate = { shareIdentity: shareIdentityValue, fullName: fullNameValue };
+export const processForm = [...validators, async (req: Request, res: Response, next: NextFunction): Promise<void> => {
```
markpit, line 25: function needs to use an await if you declare it as async, therefore you can change function signature to 
export const processForm = [...validators, (req: Request, res: Response, next: NextFunction) => {

to remove the async and Promise<void> as we are not using await

src/routes/routes.ts
```diff
@@ -28,7 +28,7 @@ const renderTemplate = (template: string) => (req: Request, res: Response) => {
 router.get("/", renderTemplate(Templates.INDEX));
 
 router.get(pageURLs.OBJECTING_ENTITY_NAME, renderTemplate(Templates.OBJECTING_ENTITY_NAME));
-router.post(pageURLs.OBJECTING_ENTITY_NAME, objectingEntityNameRoute.post);
+router.post(pageURLs.OBJECTING_ENTITY_NAME, objectingEntityNameRoute.processForm);
```
markpit, line 31: could we keep it as post for consistency

markpit (approved):

ICoxCH (approved):

PR #94, opened 2020-09-15
obj-265 ensure values are kept on error page

mattch1:
Added values to name page, so that they are held when an error is displayed.

### DIFF
--- a/src/controllers/objecting.entity.name.controller.ts
+++ b/src/controllers/objecting.entity.name.controller.ts
@@ -25,7 +25,7 @@ const validators = [
 export const post = [...validators, (req: Request, res: Response, next: NextFunction) => {
   const errors = validationResult(req);
   if (!errors.isEmpty()) {
-    return showErrorsOnScreen(errors, res);
+    return showErrorsOnScreen(errors, req, res);
   }
   const fullNameValue: string = req.body.fullName;
   const shareIdentityValue: boolean = req.body.shareIdentity === "yes";
@@ -40,7 +40,7 @@ export const post = [...validators, (req: Request, res: Response, next: NextFunc
   }
 }];
 
-const showErrorsOnScreen = (errors: Result, res: Response) => {
+const showErrorsOnScreen = (errors: Result, req: Request, res: Response) => {
   const errorListData: GovUkErrorData[] = [];
   let objectingEntityNameErr: GovUkErrorData | undefined = undefined;
   let shareIdentityErr: GovUkErrorData | undefined = undefined;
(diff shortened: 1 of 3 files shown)


src/controllers/objecting.entity.name.controller.ts
```diff
@@ -57,7 +57,12 @@ const showErrorsOnScreen = (errors: Result, res: Response) => {
       }
       errorListData.push(govUkErrorData);
     });
+
+  const fullNameValue: string = req.body.fullName
   return res.render(Templates.OBJECTING_ENTITY_NAME, {
+    fullNameValue,
```
markpit, line 63: can we add a unit test to cover this new code?

views/objecting-entity-name.html
```diff
@@ -60,11 +61,13 @@
...
               value: "yes",
-              text: "Yes"
+              text: "Yes",
+              checked: isYesChecked
             },
             {
               value: "no",
-              text: "No"
+              text: "No",
+              checked: isNoChecked
```
ICoxCH, line 70: Could this be simplified to one boolean?
isChecked, and this value be !isChecked

SVarallo:
This should be for story: https://companieshouse.atlassian.net/browse/OBJ-267 not OBJ-265. PR should be renamed to show this.

views/objecting-entity-name.html
mattch1: I think these need to be dealt with separately so as to cover the 3 senarios:
Neither selected
Yes selected
No selected.

src/controllers/objecting.entity.name.controller.ts
mattch1: Unit test coverage added.

views/objecting-entity-name.html
ICoxCH: I forgot about the Neither selected scenario, fair enough

src/controllers/objecting.entity.name.controller.ts
SVarallo: Will need a unit test to cover the retention of the name as well

SVarallo: Scratch that, it's covered

test/controller/objecting.entity.name.controller.spec.unit.ts
```diff
@@ -88,6 +88,22 @@ describe("objecting entity name tests", () => {
     expect(response.status).toEqual(200);
     expect(response.text).toContain(ENTER_FULL_NAME);
     expect(response.text).not.toContain(SELECT_TO_DIVULGE);
+    expect(response.text).toContain("value=\"yes\" checked");
```
markpit, line 91: could you add a not to contain 'no' and similar in the other test for 'yes'

mattch1: Already done.

ICoxCH: I think Mark means adding
`expect(response.text).not.toContain("value=\"no\" checked");`
In this test, and the same for 'yes' in the test below.
Checking that it is just "Yes" checked here and "No" is not checked

markpit: yes , we are checking if yes is checked, but the test will pass if both yes and no are checked

mattch1: I did know you could check both I'll have to try that out - updated the tests

markpit (approved):

ICoxCH (approved):

PR #97, opened 2020-09-17
OBJ-279 change links on webpage

mattch1:
Add change links to the check you answers web page.


views/check-your-answers.html
```diff
@@ -48,6 +48,15 @@ <h2 class="govuk-heading-m">
             },
             value: {
               text: objection.created_by.fullName
+            },
+            actions: {
+              items: [
+                {
+                  href: "/objecting-entity-name/?id=1",
```
ICoxCH, line 55: It may be better to work backwards here, is adding a query param to the url the best way to achieve what we want to?
If this param is to be used as a flag to notify the page that clicking continue takes you straight to check-your-answers instead of the normal flow, then a user could enter this url manually and bypass the normal flow completely.
Would recommend we organise a story breakdown and decide how we want to handle this before changing this page to incorporate the links

markpit, line 55: I don't think we need an id? as there will only be one name per application. if id is hardcoded to 1 then it will always be the same.

mattch1: Will remove the queries - these are really just placeholders to be dealt with as we progress this further - adding change link was an easy thing to get started with while waiting for the story breakdown.

mattch1:
Updated PR.

mattch1:
Are you referring to the CHANGE_ANSWER_KEY flag?

src/controllers/change.answers.controller.ts
```diff
@@ -0,0 +1,19 @@
...
+import {Session} from "ch-node-session-handler";
+import {STRIKE_OFF_OBJECTIONS} from "../model/page.urls";
+import {CHANGE_ANSWER_KEY} from "../constants";
+
+export const get = (req: Request, res: Response, next: NextFunction) => {
+  const session: Session | undefined = req.session as Session;
+  if (session) {
+    const query: string | undefined = req.query["changePage"] as string;
+    if (query) {
+      addToObjectionSession(session, CHANGE_ANSWER_KEY, true);
```
ICoxCH, line 12: Sorry Matt, that comment was supposed to go here, but yes, does the scope of this story include clearing this flag on the get check your answers controller

mattch1: OBJ-287 was created for this.

ICoxCH: Yep, I agree with you

ICoxCH (approved):

src/controllers/change.answers.controller.ts
```diff
@@ -0,0 +1,19 @@
+import {NextFunction, Request, Response} from "express";
+import {addToObjectionSession} from "../services/objection.session.service";
+import {Session} from "ch-node-session-handler";
+import {STRIKE_OFF_OBJECTIONS} from "../model/page.urls";
+import {CHANGE_ANSWER_KEY} from "../constants";
+
+export const get = (req: Request, res: Response, next: NextFunction) => {
+  const session: Session | undefined = req.session as Session;
+  if (session) {
+    const query: string | undefined = req.query["changePage"] as string;
```
mwestacott, line 10: Would 'changePage' be a better name for this constant/variable? 'query' is the operation we seem to be performing on the request to get the name of the page to change to...

src/controllers/change.answers.controller.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (query) {
+      addToObjectionSession(session, CHANGE_ANSWER_KEY, true);
+      const url: string = STRIKE_OFF_OBJECTIONS + "/" + query;
+      return res.redirect(url);
+    }
+    return next(new Error("No query present"));
```
mwestacott, line 16: Please see comment above - something like "No page name present to change to" would make more sense as the log message I think.

src/controllers/change.answers.controller.ts
mattch1: Chnaged the name but kept query as part of it because we test whether it is truthy or not.

src/controllers/change.answers.controller.ts
mattch1: Updated message.

src/controllers/change.answers.controller.ts
```diff
@@ -7,13 +7,13 @@ import {CHANGE_ANSWER_KEY} from "../constants";
 export const get = (req: Request, res: Response, next: NextFunction) => {
   const session: Session | undefined = req.session as Session;
   if (session) {
-    const query: string | undefined = req.query["changePage"] as string;
-    if (query) {
+    const changePageQuery: string | undefined = req.query["changePage"] as string;
```
mwestacott, line 10: @mattch1 sorry to be picky, but is this really a 'query'? It looks to me like 'query' is a function we call on the request object in order to get the value of the 'changePage' path variable. So what's returned is not a query, just a value. I think just 'changePage' as a variable name would be less misleading.

mattch1: The value is the value of the query on the url e.g. url?changeName=startPage. The "query" conveys where the value has come from so it would no be less misleading to leave it out.

mwestacott: Ok. Fair enough. Happy with the reasoning here. Will approve.

mwestacott (approved):

PR #101, opened 2020-09-22
OBJ-235 Add try-catch to async calls to api

ICoxCH:
Express 4 does not pass errors thrown in asynchronous calls to the global error handler.
This pr adds missing try/catch blocks to catch the errors and pass them to the error handler.

### DIFF
--- a/src/controllers/enter.information.controller.ts
+++ b/src/controllers/enter.information.controller.ts
@@ -26,7 +26,11 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
 
   const reason: string = req.body.information;
 
-  await updateObjectionReason(company.companyNumber, objectionId, token, reason);
+  try {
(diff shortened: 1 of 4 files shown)


mattch1 (approved):

src/controllers/enter.information.controller.ts
```diff
@@ -26,7 +26,11 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
 
   const reason: string = req.body.information;
 
-  await updateObjectionReason(company.companyNumber, objectionId, token, reason);
+  try {
+    await updateObjectionReason(company.companyNumber, objectionId, token, reason);
+  } catch (e) {
+    return next(e);
```
markpit, line 32: could we add unit tests to make sure these fixes aren't reverted accidentally

markpit (approved):

PR #102, opened 2020-09-22
Feature/obj 281 name page preload data

mattch1:
Check that answer flag is not undefined and not false and only if true will we delete the flag ( temp code here to stop details appear on page even when they should not ) and then get the details and pass them to the template.

### DIFF
--- a/src/controllers/objecting.entity.name.controller.ts
+++ b/src/controllers/objecting.entity.name.controller.ts
@@ -12,7 +12,7 @@ import { Session } from "ch-node-session-handler";
 import { Templates } from "../model/template.paths";
 import { getObjection } from "../services/objection.service";
 import logger from "../utils/logger";
-import {CHANGE_ANSWER_KEY, SESSION_OBJECTION_CREATE} from "../constants";
+import { CHANGE_ANSWER_KEY, SESSION_OBJECTION_CREATE } from "../constants";
 
 const FULL_NAME_FIELD = "fullName";
 const DIVULGE_INFO_FIELD = "shareIdentity";
--- a/src/controllers/objecting.entity.name.controller.ts
+++ b/src/controllers/objecting.entity.name.controller.ts
@@ -73,8 +73,6 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
   const session: Session | undefined = req.session as Session;
   if (session) {
     if (retrieveFromObjectionSession(session, CHANGE_ANSWER_KEY)) {
-      // TODO OBJ-287 handle this more formally.
-      delete retrieveObjectionSessionFromSession(session)[CHANGE_ANSWER_KEY];
       return await showPageWithMongoData(session, res, next);
     } else {
       return showPageWithSessionDataIfPresent(session, res);
--- a/src/controllers/objecting.entity.name.controller.ts
+++ b/src/controllers/objecting.entity.name.controller.ts
@@ -12,7 +12,7 @@ import { Session } from "ch-node-session-handler";
 import { Templates } from "../model/template.paths";
 import { getObjection } from "../services/objection.service";
 import logger from "../utils/logger";
-import {CHANGE_ANSWER_KEY, SESSION_OBJECTION_CREATE} from "../constants";
+import { CHANGE_ANSWER_KEY } from "../constants";
 
 const FULL_NAME_FIELD = "fullName";
 const DIVULGE_INFO_FIELD = "shareIdentity";
@@ -43,6 +43,14 @@ const getExistingPageData = async (session: Session, res: Response, next: NextFu
   }
 }
 
+const isChangeAnswerFlagPresentSetAndSetToTrue = (session: Session): boolean => {
+  const changeAnswer: boolean = retrieveFromObjectionSession(session, CHANGE_ANSWER_KEY);
+  if (changeAnswer) {
+    return changeAnswer === true;
+  }
+  return false;
+};
+
 /**
  * GET checks for change flag and renders page
  * @param req
@@ -52,9 +60,7 @@ const getExistingPageData = async (session: Session, res: Response, next: NextFu
 export const get = async (req: Request, res: Response, next: NextFunction) => {
   const session: Session | undefined = req.session as Session;
   if (session) {
-    if (retrieveObjectionSessionFromSession(session) &&
-        retrieveFromObjectionSession(session, CHANGE_ANSWER_KEY)) {
-
+    if (isChangeAnswerFlagPresentSetAndSetToTrue(session)) {
       // TODO OBJ-287 handle this more formally.
       delete retrieveObjectionSessionFromSession(session)[CHANGE_ANSWER_KEY];
 
--- a/src/controllers/objecting.entity.name.controller.ts
+++ b/src/controllers/objecting.entity.name.controller.ts
@@ -30,8 +30,8 @@ const getSessionPageDataIfPresent = (session: Session, res: Response) => {
   if (objectionCreate) {
     existingName = objectionCreate.fullName;
     const existingSharedIdentity = objectionCreate.shareIdentity;
-    yesChecked = existingSharedIdentity === true;
-    noChecked = existingSharedIdentity !== true;
+    yesChecked = existingSharedIdentity;
+    noChecked = !existingSharedIdentity;
   }
 
   return res.render(Templates.OBJECTING_ENTITY_NAME, {
@@ -47,7 +47,7 @@ const getExistingPageData = async (session: Session, res: Response, next: NextFu
     const objection: Objection = await getObjection(session) as Objection;
     const existingName: string = objection.created_by.fullName;
     const existingShareIdentity: boolean = objection.created_by.shareIdentity;
-    if (existingName && existingShareIdentity) {
+    if (existingName && existingShareIdentity !== undefined) {
       return res.render(Templates.OBJECTING_ENTITY_NAME, {
         fullNameValue: existingName,
         isYesChecked: existingShareIdentity,
--- a/src/controllers/objecting.entity.name.controller.ts
+++ b/src/controllers/objecting.entity.name.controller.ts
@@ -3,10 +3,16 @@ import { check, Result, ValidationError, validationResult } from "express-valida
 import { ErrorMessages } from "../model/error.messages";
 import { createGovUkErrorData, GovUkErrorData } from "../model/govuk.error.data";
 import { OBJECTIONS_COMPANY_NUMBER } from "../model/page.urls";
-import { addObjectionCreateToObjectionSession } from "../services/objection.session.service";
-import { ObjectionCreate } from "../modules/sdk/objections";
+import {
+  addObjectionCreateToObjectionSession, retrieveFromObjectionSession,
+  retrieveObjectionSessionFromSession
+} from "../services/objection.session.service";
+import { Objection, ObjectionCreate } from "../modules/sdk/objections";
 import { Session } from "ch-node-session-handler";
 import { Templates } from "../model/template.paths";
+import { getObjection } from "../services/objection.service";
+import logger from "../utils/logger";
+import {CHANGE_ANSWER_KEY, SESSION_OBJECTION_CREATE} from "../constants";
 
 const FULL_NAME_FIELD = "fullName";
 const DIVULGE_INFO_FIELD = "shareIdentity";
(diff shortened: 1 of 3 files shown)


src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,58 @@ const validators = [
   check(DIVULGE_INFO_FIELD).not().isEmpty().withMessage(ErrorMessages.SELECT_TO_DIVULGE),
 ];
 
+const getExistingPageData = async (session: Session, res: Response, next: NextFunction) => {
+  try {
+    const objection: Objection = await getObjection(session) as Objection;
+    const existingName: string = objection.created_by.fullName;
+    const existingShareIdneity: string = (objection.created_by.shareIdentity === true) ? "yes" : "no";
```
ICoxCH, line 29: Typo in var name

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,58 @@ const validators = [
...
+    if (existingName && existingShareIdneity) {
+      return res.render(Templates.OBJECTING_ENTITY_NAME, {
+        fullNameValue: existingName,
+        isYesChecked: existingShareIdneity === "yes",
```
ICoxCH, line 33: why are we transforming it from a boolean to "yes"/"no" above then transforming it back to a boolean here?
can we just use objection.created_by.shareIdentity?

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,58 @@ const validators = [
...
+  } catch (e) {
+    logger.error(e.message);
+    return next(e);
+  }
+}
+
+const isChangeAnswerFlagPresentSetAndSetToTrue = (session: Session): boolean => {
+  const changeAnswer: boolean = retrieveFromObjectionSession(session, CHANGE_ANSWER_KEY);
+  if (changeAnswer) {
+    return changeAnswer === true;
```
ICoxCH, line 49: change answer is a boolean, can we just return changeAnswer? as retrieveFromObjectionSession should either return a boolean or error.

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,58 @@ const validators = [
   check(DIVULGE_INFO_FIELD).not().isEmpty().withMessage(ErrorMessages.SELECT_TO_DIVULGE),
 ];
 
+const getExistingPageData = async (session: Session, res: Response, next: NextFunction) => {
+  try {
+    const objection: Objection = await getObjection(session) as Objection;
+    const existingName: string = objection.created_by.fullName;
+    const existingShareIdneity: string = (objection.created_by.shareIdentity === true) ? "yes" : "no";
+    if (existingName && existingShareIdneity) {
+      return res.render(Templates.OBJECTING_ENTITY_NAME, {
```
ICoxCH, line 31: we will also need to pass in a way for the template to know that upon clicking submit, to go straight back to the check your answers page won't we?

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,58 @@ const validators = [
...
+ * @param req
+ * @param res
+ * @param next
+ */
+export const get = async (req: Request, res: Response, next: NextFunction) => {
+  const session: Session | undefined = req.session as Session;
+  if (session) {
+    if (isChangeAnswerFlagPresentSetAndSetToTrue(session)) {
+      // TODO OBJ-287 handle this more formally.
+      delete retrieveObjectionSessionFromSession(session)[CHANGE_ANSWER_KEY];
```
ICoxCH, line 65: I believe clearing the flag will be done on loading the check your answers page, not on the indiviudal pages

src/controllers/objecting.entity.name.controller.ts
ICoxCH: I also believe objection.created_by.shareIdentity === true can be simplified to objection.created_by.shareIdentity

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,58 @@ const validators = [
...
+    const existingShareIdneity: string = (objection.created_by.shareIdentity === true) ? "yes" : "no";
+    if (existingName && existingShareIdneity) {
+      return res.render(Templates.OBJECTING_ENTITY_NAME, {
+        fullNameValue: existingName,
+        isYesChecked: existingShareIdneity === "yes",
+        isNoChecked: existingShareIdneity === "no",
+        templateName: Templates.OBJECTING_ENTITY_NAME,
+      });
+    } else {
+      return next(new Error("Existing data not present"));
```
markpit, line 38: we need to show the page with no data, as when a customer is completing the screens first time through, the data will not have been submitted. I haven't tested it but it looks like line 38 will show the error screen if there is no existing data.

src/controllers/objecting.entity.name.controller.ts
mattch1: Could be, but I'm taking account of undefined as well as false being returned the "===" checks that it is boolean and not undefined.

src/controllers/objecting.entity.name.controller.ts
markpit: sorry, ignore above, i can see further down that it renders page normally

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,58 @@ const validators = [
...
+    }
+  } catch (e) {
+    logger.error(e.message);
+    return next(e);
+  }
+}
+
+const isChangeAnswerFlagPresentSetAndSetToTrue = (session: Session): boolean => {
```
markpit, line 46: function name looks like it has 1 too many 'set's

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,58 @@ const validators = [
...
+ * @param next
+ */
+export const get = async (req: Request, res: Response, next: NextFunction) => {
+  const session: Session | undefined = req.session as Session;
+  if (session) {
+    if (isChangeAnswerFlagPresentSetAndSetToTrue(session)) {
+      // TODO OBJ-287 handle this more formally.
+      delete retrieveObjectionSessionFromSession(session)[CHANGE_ANSWER_KEY];
+
+      return await getExistingPageData(session, res, next);
```
markpit, line 67: we'd want to show any data they've entered regardless of whether the change flag is present. The change flag should only be used to change the nest page to go to. This way, data will be shown if they use the  'back' button as well.

test/controller/objecting.entity.name.controller.spec.unit.ts
```diff
@@ -33,25 +36,107 @@ const mockAuthenticationMiddleware = authenticationMiddleware as jest.Mock;
...
+      }
+
+      return next(new Error("No session on request"));
+    });
+  });
+
+  it("should render empty objecting entity name page if no change answer flag is set in session", async() => {
+    mockRetrieveFromObjectionSession.mockReset();
+    mockRetrieveFromObjectionSession.mockReturnValueOnce(undefined);
+    const response = await request(app).get(OBJECTIONS_OBJECTING_ENTITY_NAME)
```
markpit, line 72: could do with some whitespace to separate mocks from the 'get' call and from the expects.

src/controllers/objecting.entity.name.controller.ts
mattch1: I think we want to minimise the time it is left in the session as we don't want there to be the details to be on the page unexpectedly - the details for theis will be worked out for case OBJ-287.

src/controllers/objecting.entity.name.controller.ts
mattch1: It will this is when we expect there to be data and there is none - line 61 deals with completing the screens first time through. However we could just redner the page here as well and put an error in the logs - any thoughts?

src/controllers/objecting.entity.name.controller.ts
mattch1: That was deliberate to say the flag is a) set and b) set to the value true, but "isChangeAnswerFlagPresentAndSetToTrue" should suffice.

mattch1: Actually will not be using this function anymore refactoring for another comment.

src/controllers/objecting.entity.name.controller.ts
mattch1: Could be tricky as we cache this in the session initally (pre check company details submit) so getting the objection will return nothing. I thought the aim of this ticket was to base this on the change flag.

src/controllers/objecting.entity.name.controller.ts
ICoxCH: neither created_by or share identity are optional fields in an Objection, so surely it is impossible for it to be null here? (Maybe being Node it is possible)
I would of thought if there is no CreatedBy the getObjection call would error on missing those fields.

src/controllers/objecting.entity.name.controller.ts
ICoxCH: oh ok, I would of added that field here, and modified the template in the other ticket, but either works.

src/controllers/objecting.entity.name.controller.ts
mattch1: Updating this just to look for truthy or falsy (which covers false and undefined).

test/controller/objecting.entity.name.controller.spec.unit.ts
mattch1: Added spaces.

src/controllers/objecting.entity.name.controller.ts
mattch1: Added another function for this.

mattch1:
Objection.entity.name.controller test: 

Have a failing jest unit test - trying to call

await getObjection(session) as Objection in the controller code.
(this is objection service not sdk) 

and it comes back undefined - but in the test:

mockGetObjection.mockReset().mockResolvedValueOnce(mockObjection);
It is as though this line never happened - the result is the same if I remove it I cannot work out why this is being ignored.

src/controllers/objecting.entity.name.controller.ts
mattch1: Removed conversion to yes/no

src/controllers/objecting.entity.name.controller.ts
mattch1: Removed function just use the function call

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -3,10 +3,16 @@ import { check, Result, ValidationError, validationResult } from "express-valida
...
+import {
+  addObjectionCreateToObjectionSession, retrieveFromObjectionSession,
+  retrieveObjectionSessionFromSession
+} from "../services/objection.session.service";
+import { Objection, ObjectionCreate } from "../modules/sdk/objections";
 import { Session } from "ch-node-session-handler";
 import { Templates } from "../model/template.paths";
+import { getObjection } from "../services/objection.service";
+import logger from "../utils/logger";
+import {CHANGE_ANSWER_KEY, SESSION_OBJECTION_CREATE} from "../constants";
```
markpit, line 15: spaces needed around  {} on line 15

mattch1: Added spaces.

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,67 @@ const validators = [
...
+    fullNameValue: existingName,
+    isYesChecked: yesChecked,
+    isNoChecked: noChecked,
+    templateName: Templates.OBJECTING_ENTITY_NAME,
+  });
+}
+
+const getExistingPageData = async (session: Session, res: Response, next: NextFunction) => {
+  try {
+    const objection: Objection = await getObjection(session) as Objection;
```
ICoxCH, line 47: I don't believe we need "as Objection" here, getObjection returns an Objection, and you have set the type to Objection without it.

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,67 @@ const validators = [
   check(DIVULGE_INFO_FIELD).not().isEmpty().withMessage(ErrorMessages.SELECT_TO_DIVULGE),
 ];
 
+const getSessionPageDataIfPresent = (session: Session, res: Response) => {
```
markpit, line 25: wondering if we could improve the 2 new function names - the getSessionPageData... makes me think the function is going to return some data from the session we can use, but in reality it is showing the page with the session data. So, how about naming it showPageWithSessionDataIfPresent? - similar for getExistingPageData -> showPageWithMongoData perhaps?

src/controllers/objecting.entity.name.controller.ts
ICoxCH: If this flag is cleared on the GET, then it won't be present on the POST to decide where to send the user next.
Don't worry about clearing the flag in this pr, 287 willl ensure the flag gets cleared when it needs to

src/controllers/objecting.entity.name.controller.ts
mattch1: Remove dunecessary typing

src/controllers/objecting.entity.name.controller.ts
mattch1: Names changed.

src/controllers/objecting.entity.name.controller.ts
markpit: Not sure if it's just my browser but the spaces aren't showing for me, still seeing
import {CHANGE_ANSWER_KEY, SESSION_OBJECTION_CREATE}

mattch1: See commit just below.

mattch1: There were two instances - both now up to date.

src/controllers/objecting.entity.name.controller.ts
markpit: I think Ian is right, if we delete the change answer key here, it won't be available when 'post' is called and we'll end up going to the next page instead of back to the 'change your answers' page

mattch1: Alright removed this - it was only mean to be a stop gap - it will mean that we will have the details appearing on the page for the lifecycle of the session until 287 is completed.

markpit (approved):

ICoxCH (approved):

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -16,6 +22,65 @@ const validators = [
...
+ */
+export const get = async (req: Request, res: Response, next: NextFunction) => {
+  const session: Session | undefined = req.session as Session;
+  if (session) {
+    if (retrieveFromObjectionSession(session, CHANGE_ANSWER_KEY)) {
+      return await showPageWithMongoData(session, res, next);
+    } else {
+      return showPageWithSessionDataIfPresent(session, res);
+    }
+    return next(new Error("No Session present"));
```
markpit, line 80: just spotted a bug i think, line 80 the return error for no session is inside the if(session) block meaning the session is present, and it would never be called anyway due to the if/else above returning the pages. I think line 80 should go between 81 and 82.

PR #104, opened 2020-09-23
OBJ-287 remove change flag when loading check answers page

ICoxCH:


mwestacott (approved):

src/services/objection.session.service.ts
```diff
@@ -107,6 +107,15 @@ export const retrieveFromObjectionSession = (session: Session, key: string): any
   return objectionsExtraData[key];
 };
 
+export const deleteFromObjectionSession = (session: Session, key: string): void => {
```
markpit, line 110: could we add a test into objection.session.service.spec.unit.ts just to test this function? looks like it should be simple-ish

markpit (approved):

PR #105, opened 2020-09-24
OBJ-284 continue to check answers page

mattch1:
Set redirect to check answers when the flag is set.
Also when the flag is set, the service is called to update the user details.
Will need to update the api to accept the changes and add them to mongo.

Corrected the session error for the get function (OBJ-281)

### DIFF
--- a/src/controllers/company.number.controller.ts
+++ b/src/controllers/company.number.controller.ts
@@ -8,6 +8,50 @@ import {
   retrieveAccessTokenFromSession,
 } from "../services/objection.session.service";
 import logger from "../utils/logger";
+import { check, validationResult } from "express-validator";
+import { CompanySearchErrorMessages } from "../model/error.messages";
+import { createGovUkErrorData, GovUkErrorData } from "../model/govuk.error.data";
+import { Templates } from "../model/template.paths";
+import { ValidationError } from "../model/validation.error";
+
+const companyNumberFieldName: string = "companyNumber";
+
+// validator middleware that checks for an empty company number or one that is too long
+const preValidators = [
+  check(companyNumberFieldName).blacklist(" ").escape().not().isEmpty().withMessage(CompanySearchErrorMessages.NO_COMPANY_NUMBER_SUPPLIED),
+  check(companyNumberFieldName).blacklist(" ").escape().isLength({max: 8}).withMessage(CompanySearchErrorMessages.COMPANY_NUMBER_TOO_LONG),
+];
+
+// pads company number to 8 digits with zeros and removes whitespace
+const padCompanyNumber = (req: Request, res: Response, next: NextFunction): void => {
+  let companyNumber: string = req.body.companyNumber;
+  if (/^([a-zA-Z]{2}?)/gm.test(companyNumber)) {
+    companyNumber = formatCompanyNumber(companyNumber, 2);
+  } else if (/^([a-zA-Z])/gm.test(companyNumber)) {
+    companyNumber = formatCompanyNumber(companyNumber, 1);
+  } else {
+    companyNumber = companyNumber.padStart(8, "0");
+  }
+  req.body.companyNumber = companyNumber;
+  return next();
+};
+
+const formatCompanyNumber = (companyNumber: string, leadPoint: number): string  => {
+  const leadingLetters = companyNumber.substring(0, leadPoint);
+  let trailingChars = companyNumber.substring(leadPoint, companyNumber.length);
+  trailingChars = trailingChars.padStart((8 - leadPoint), "0");
+  return leadingLetters + trailingChars;
+};
+
+// validator middleware that checks for invalid characters in the input
+const postValidators = [
+  check(companyNumberFieldName).blacklist(" ").escape().custom((value: string) => {
+    if (!/^[0-9]{8}$|^([a-zA-Z])[0-9]{7}$|^([a-zA-Z]{2})[0-9]{6}$/gm.test(value)) {
+      throw new Error(CompanySearchErrorMessages.INVALID_COMPANY_NUMBER);
+    }
+    return true;
+  }),
+];
 
 /**
  * POST controller for company number screen
@@ -15,8 +59,19 @@ import logger from "../utils/logger";
  * @param res
  * @param next
  */
-export const post = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+const post = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
   logger.debugRequest(req, "Attempt to retrieve and show the company details");
+
+  const errors = validationResult(req);
+
+  // render errors in the view
+  if (!errors.isEmpty()) {
+    const errorText = errors.array({ onlyFirstError: true })
+      .map((err: ValidationError) => err.msg)
+      .pop() as string;
+
+    return buildError(res, errorText);
+  }
 
   const companyNumber: string = req.body.companyNumber;
   const session: Session = req.session as Session;
@@ -31,6 +86,27 @@ export const post = async (req: Request, res: Response, next: NextFunction): Pro
     addCompanyProfileToObjectionSession(session, company);
     return res.redirect(OBJECTIONS_CONFIRM_COMPANY);
   } catch (e) {
-    return next(e);
+    logger.error(`Error fetching company profile for company number ${companyNumber}`);
+
+    if (e.status === 404) {
+      buildError(res, CompanySearchErrorMessages.COMPANY_NOT_FOUND);
+    } else {
+      return next(e);
+    }
   }
 };
+
+const buildError = (res: Response, errorMessage: string): void => {
+  const companyNumberErrorData: GovUkErrorData = createGovUkErrorData(
+    errorMessage,
+    "#company-number",
+    true,
+    "");
+  return res.render(Templates.COMPANY_NUMBER, {
+    companyNumberErr: companyNumberErrorData,
+    errorList: [companyNumberErrorData],
+    templateName: Templates.COMPANY_NUMBER,
+  });
+};
+
+export default [...preValidators, padCompanyNumber, ...postValidators, post];
--- a/src/controllers/objecting.entity.name.controller.ts
+++ b/src/controllers/objecting.entity.name.controller.ts
@@ -3,10 +3,16 @@ import { check, Result, ValidationError, validationResult } from "express-valida
 import { ErrorMessages } from "../model/error.messages";
 import { createGovUkErrorData, GovUkErrorData } from "../model/govuk.error.data";
 import { OBJECTIONS_COMPANY_NUMBER } from "../model/page.urls";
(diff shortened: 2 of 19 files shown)


src/modules/sdk/objections/types.ts
```diff
@@ -37,6 +37,8 @@ export interface ObjectionCreate {
  * @interface
  */
 export interface ObjectionPatch {
+  fullName?: string;
```
ICoxCH, line 40: json fields should be snake_case

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -2,17 +2,20 @@ import { NextFunction, Request, Response } from "express";
...
+  addObjectionCreateToObjectionSession,
+  retrieveAccessTokenFromSession,
+  retrieveCompanyProfileFromObjectionSession,
+  retrieveFromObjectionSession,
 } from "../services/objection.session.service";
 import { Objection, ObjectionCreate } from "../modules/sdk/objections";
 import { Session } from "ch-node-session-handler";
 import { Templates } from "../model/template.paths";
-import { getObjection } from "../services/objection.service";
+import {getObjection, updateObjectionReason, updateObjectionUserDetails} from "../services/objection.service";
```
ICoxCH, line 15: spaces

test/controller/objecting.entity.name.controller.spec.unit.ts
```diff
@@ -177,7 +180,10 @@ describe("objecting entity name tests", () => {
     expect(response.status).toEqual(500);
   });
 
-  it("should render the company number page when posting with entered details", async () => {
+  it("should render the company number page when posting with entered details no change answers", async () => {
```
ICoxCH, line 183: this could possibly be named clearer.
Maybe "should render the company number page when posting with entered details and change answer flag not present"

test/controller/objecting.entity.name.controller.spec.unit.ts
```diff
@@ -188,6 +194,29 @@ describe("objecting entity name tests", () => {
 
     expect(response.status).toEqual(302);
     expect(response.header.location).toEqual(OBJECTIONS_COMPANY_NUMBER);
+    expect(mockRetrieveFromObjectionSession).toHaveBeenCalledTimes(1);
+    expect(mockRetrieveAccessToken).not.toBeCalled();
+    expect(mockRetrieveCompanyProfileFromSession).not.toBeCalled();
+  });
+
+  it("should render the company number page when posting with entered details change answers false", async () => {
```
ICoxCH, line 202: similarly to above, could be named "should render the company number page when posting with entered details and change answers flag set to false"

test/controller/objecting.entity.name.controller.spec.unit.ts
```diff
@@ -264,6 +293,31 @@ describe("objecting entity name tests", () => {
     expect(response.status).toEqual(500);
     expect(response.text).toContain(ERROR_500);
   });
+
+  it("should navigate to check you answers and send details to mongo when check answers", async () => {
```
ICoxCH, line 297: Perhaps: "should navigate to check you answers and send details to mongo when check answers flag set to true"

src/modules/sdk/objections/types.ts
mattch1: I agree. There are a number of other fields in camel case, so what I'd like to do is raise a bug for this and defer to that as there is other work depending on these fields at present and it will need to be tersted thoroughly.

src/controllers/objecting.entity.name.controller.ts
mattch1: Put spaces in.

test/controller/objecting.entity.name.controller.spec.unit.ts
mattch1: Renamed and the two below - abbreviated slightly.

src/controllers/objecting.entity.name.controller.ts
markpit: spaces still missing for me, i think your last commit was just the unit test

markpit: the updateObjectionReason is showing as an unused import in GoLand. You can check for errors in your IDE before committing, they will be highlighted in different colours to make them stand out. We should also have eslint rules for unused imports and bracket spaces.

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -77,7 +80,20 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
     } else {
       return showPageWithSessionDataIfPresent(session, res);
     }
-    return next(new Error("No Session present"));
+  }
+  return next(new Error("No Session present"));
```
markpit, line 84: we need a test to check the error screen is shown for this line, if we had one it would have failed and highlighted the bug. We need to add a test to prevent the bug from being re-introduced.

src/controllers/objecting.entity.name.controller.ts
```diff
@@ -77,7 +80,20 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
...
+const updateMongoWithChangedUserDetails = async (session: Session,
+                                                 objectionCreate: ObjectionCreate,
+                                                 next: NextFunction) => {
+  try {
+    const company: ObjectionCompanyProfile = retrieveCompanyProfileFromObjectionSession(session);
+    const objectionId: string = retrieveFromObjectionSession(session, SESSION_OBJECTION_ID);
+    const token: string = retrieveAccessTokenFromSession(session);
+    return await updateObjectionUserDetails(company.companyNumber, objectionId, token, objectionCreate);
+  } catch (e) {
+    return next(e);
```
markpit, line 96: need a test to check the error screen is shown for line 96

src/services/objection.service.ts
```diff
@@ -31,6 +31,13 @@ export const createNewObjection = async (companyNumber: string, token: string, c
   return objectionId;
 };
 
+export const updateObjectionUserDetails = async (companyNumber: string, objectionId: string,
```
markpit, line 34: All new lines of code we add should be tested with a unit test to ensure it functions properly. This new function will need a new unit test in objection.service.spec.unit.ts.

src/controllers/objecting.entity.name.controller.ts
mattch1: There is a test for this but as the error hadling is a catch all at the end of the chain it is difficult discern whether it reroutes to the error screen and returns 500 because of any one particular error. If you take out the next line with the "No session present" message then the result is the same.

src/controllers/objecting.entity.name.controller.ts
mattch1: Added a test removed the return statement.

src/services/objection.service.ts
mattch1: Put a test in here to make sure the sdk is being called with the values provided.

src/controllers/objecting.entity.name.controller.ts
markpit: my 'npm run coverage' command is showing line 84 as being un-tested. We just want to ensure that the error page is displayed if we call 'get' with an undefined or null session. If we didn't have the return next(new err..), the browser request would time out as there is no session so nothing would be returned. So, essentially:
**When** the 'get' function is called
**And** the session is missing
**Then** the error screen should be returned

we should be able to look at the text returned in the response to see if it is the error screen.

mattch1: Yes yes yes yes I understand that that's not the problem here.

markpit: ok, i'll have to look into it a bit later, see if i can sort it

mattch1: A test was written to cover this essential, crucial and very important fact just stated, yet in spite of it all that line is not covered regardless of whether I set the mock not to initailise the session or explicitly set it to undefined.

mattch1: Children, animal and node unit tests hey.

markpit: ah ok, just thought it hadn't been covered. The tests aren't the easiest, but practice makes perfect !

src/controllers/company.number.controller.ts
```diff
@@ -8,7 +8,7 @@ import {
   retrieveAccessTokenFromSession,
 } from "../services/objection.session.service";
 import logger from "../utils/logger";
-import { check, validationResult } from "express-validator/check";
+import { check, validationResult } from "express-validator";
```
markpit, line 11: warning in console when running tests that express-validator/check was deprecated

src/controllers/company.number.controller.ts
```diff
@@ -27,7 +27,7 @@ const padCompanyNumber = (req: Request, res: Response, next: NextFunction): void
   let companyNumber: string = req.body.companyNumber;
   if (/^([a-zA-Z]{2}?)/gm.test(companyNumber)) {
     companyNumber = formatCompanyNumber(companyNumber, 2);
-  } else if (/^([a-zA-Z]{1}?)/gm.test(companyNumber)) {
+  } else if (/^([a-zA-Z])/gm.test(companyNumber)) {
```
markpit, line 30: IDE highlighted that repetition {1} can be removed

src/controllers/company.number.controller.ts
```diff
@@ -46,7 +46,7 @@ const formatCompanyNumber = (companyNumber: string, leadPoint: number): string
 // validator middleware that checks for invalid characters in the input
 const postValidators = [
   check(companyNumberFieldName).blacklist(" ").escape().custom((value: string) => {
-    if (!/^[0-9]{8}$|^([a-zA-Z]{1})[0-9]{7}$|^([a-zA-Z]{2})[0-9]{6}$/gm.test(value)) {
+    if (!/^[0-9]{8}$|^([a-zA-Z])[0-9]{7}$|^([a-zA-Z]{2})[0-9]{6}$/gm.test(value)) {
```
markpit, line 49: IDE highlighted that repetition {1} can be removed

src/modules/sdk/objections/types.ts
```diff
@@ -55,11 +57,11 @@ export interface Attachment {
  * @interface
  */
 export interface Objection {
-  reason: string;
-  created_by: CreatedBy,
   attachments: Array<{
     name: string;
-  }>;
+  }>,
```
markpit, line 62: just making them alphabetically ordered

test/modules/sdk/objections/service.spec.unit.ts
```diff
@@ -38,6 +38,10 @@ const dummyObjection: Objection = {
     {
       name: "document.pdf",
     }],
+  created_by: {
```
markpit, line 41: IDE highlighted missing created_by so added it

markpit (approved):

ICoxCH (approved):

PR #120, opened 2020-10-01
OBJ-302 Simplified regex to be less vulnerable to reDos attack

ICoxCH:


src/controllers/company.number.controller.ts
```diff
@@ -47,7 +47,7 @@ const formatCompanyNumber = (companyNumber: string, leadPoint: number): string
 // validator middleware that checks for invalid characters in the input
 const postValidators = [
   check(companyNumberFieldName).blacklist(" ").escape().custom((value: string) => {
-    if (!/^[0-9]{8}$|^([a-zA-Z])[0-9]{7}$|^([a-zA-Z]{2})[0-9]{6}$/gm.test(value)) {
+    if (!/^(?=[a-z0-9]{8}$)[a-z]{0,2}\d+$/gmi.test(value)) {
```
markpit, line 50: just given it a try in https://regex101.com/r/0nFxnL/2
not sure if i have entered it wrong (as it defaults to /gm on end) but uppercase letter didn't match
d1234567 - ok
D1234567 - no match

markpit: unless we are changing case in the code?

mwestacott (approved):

src/controllers/company.number.controller.ts
ICoxCH: This has gmi on the end, so it should be case insensitive

markpit (approved):

src/controllers/company.number.controller.ts
mwestacott: Interesting. Node unit tests use upper-case letters so maybe case is changed in code. Perhaps worth adding tests with lower-case letters too though.

mwestacott (approved):

PR #123, opened 2020-10-01
OBJ-302 Updating company number regex for security vulnerability

markpit:
Updating company number regex based on Sonar report to prevent a security vulnerability

### DIFF
--- a/src/controllers/company.number.controller.ts
+++ b/src/controllers/company.number.controller.ts
@@ -47,7 +47,7 @@ const formatCompanyNumber = (companyNumber: string, leadPoint: number): string  
 // validator middleware that checks for invalid characters in the input
 const postValidators = [
   check(companyNumberFieldName).blacklist(" ").escape().custom((value: string) => {
-    if (value.length !== 8 || !/^[a-z]{0,2}?\d+?$/gmi.test(value)) {
+    if (value.length !== 8 || !/^(?:[a-z]|[a-z][a-z])?\d{6,8}?$/gmi.test(value)) {
       throw new Error(CompanySearchErrorMessages.INVALID_COMPANY_NUMBER);
     }
     return true;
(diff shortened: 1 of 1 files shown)


src/controllers/company.number.controller.ts
```diff
@@ -47,7 +47,7 @@ const formatCompanyNumber = (companyNumber: string, leadPoint: number): string
 // validator middleware that checks for invalid characters in the input
 const postValidators = [
   check(companyNumberFieldName).blacklist(" ").escape().custom((value: string) => {
-    if (value.length !== 8 || !/^[a-z]{0,2}?\d+?$/gmi.test(value)) {
+    if (value.length !== 8 || !/^(?:[a-z]|[a-z][a-z])?\d{6,8}?$/gmi.test(value)) {
```
markpit, line 50: ?:  is a non capturing group, without it the group () will be captured so we can use it later on, but we don't need to capture it, hence added ?:

markpit: Sonar doesn't like 2 or more of +*{ so i've made the 1 or 2 letters at start literal rather then use the {0,2} occurrences, so we can use {} or + after the d. I've gone for {6,8} to narrow it down a bit, + would allow infinite number of digits. We have a length check in the code before this, so any string being evaluated will always be length of 8.

mwestacott (approved):
Let's give it a whirl!

ICoxCH (approved):

ICoxCH (approved):

markpit:
made another update to the /gmi flags
remove g (global) as this would find all matches in the string instead of returning after 1st match found. This meant if we entered AB26366464646464464646 it would take 40+ steps to process it even after the first part of the string matched.

removed m for multiline mode, we only have a single line so don't need this.

PR #125, opened 2020-10-02
OBJ-321 Add filing history call to service sdk

ICoxCH:
This method in service sdk currently not called, but tested and works. Ready for use when we need to retrieve the date

### DIFF
--- a/src/services/company.filing.history.service.ts
+++ b/src/services/company.filing.history.service.ts
@@ -2,6 +2,7 @@ import logger from "../utils/logger";
 import { createApiClient } from "ch-sdk-node";
 import Resource from "ch-sdk-node/dist/services/resource";
 import { CompanyFilingHistory } from "ch-sdk-node/dist/services/company-filing-history";
+import { inspect } from "util";
 
 export const getCompanyFilingHistory = async (companyNumber: string, category: string, token: string):
   Promise<CompanyFilingHistory> => {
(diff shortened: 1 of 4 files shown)


src/services/company.filing.history.service.ts
```diff
@@ -0,0 +1,24 @@
...
+  const sdkResponse: Resource<CompanyFilingHistory> =
+    await api.companyFilingHistory.getCompanyFilingHistory(companyNumber.toUpperCase(), category);
+
+  if (sdkResponse.httpStatusCode >= 400) {
+    throw {
+      status: sdkResponse.httpStatusCode,
+    };
+  }
+
+  logger.debug("Data from company filing history SDK call " + JSON.stringify(sdkResponse, null, 2));
```
markpit, line 21: is it worth doing that trick you used last time for self references that can cause JSON.stringify to fail? Just thinking if someone in another team changed the sdk in future it may contain a self reference and break our code.

test/services/company.filing.history.service.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    expect(mockGetCompanyFilingHistory).toBeCalledWith("SC100079");
+  });
+
+  it("returns a CompanyFilingHistory object", async () => {
+    mockGetCompanyFilingHistory.mockResolvedValueOnce(dummySDKResponse);
+    const companyFilingHistory = await getCompanyFilingHistory("12345678", "category", ACCESS_TOKEN);
+    expect(companyFilingHistory).toEqual(dummySDKResponse.resource);
+  });
+});
+
```
markpit, line 32: missing a test for it throwing an error if http status code >= 400

markpit (approved):

mwestacott (approved):

PR #126, opened 2020-10-02
Obj 321 extract gaz1 date from api response

ICoxCH:
Checks if GAZ1 date should be displayed and if yes makes a call to get it, extract it from the response and display. If not, static text is displayed on the 'confirm' screen.

### DIFF
--- a/src/controllers/confirm.company.controller.ts
+++ b/src/controllers/confirm.company.controller.ts
@@ -14,6 +14,7 @@ import {
   retrieveObjectionCreateFromObjectionSession
 } from "../services/objection.session.service";
 import logger from "../utils/logger";
+import { formatCHSDateForDisplay } from "../utils/date.formatter";
 import { getCompanyFilingHistory } from "../services/company.filing.history.service";
 import { CompanyFilingHistory, FilingHistoryItem } from "ch-sdk-node/dist/services/company-filing-history";
 
@@ -31,12 +32,20 @@ const INELIGIBLE_PAGES = {
  * @param next
  */
 
-export const get = (req: Request, res: Response, next: NextFunction) => {
+export const get = async (req: Request, res: Response, next: NextFunction) => {
   if (req.session) {
     try {
       const company: ObjectionCompanyProfile = retrieveCompanyProfileFromObjectionSession(req.session);
+
+      // TODO Display 'No notice in the Gazette' text if company action code is not eligible (OBJ-324 must be completed)
+      const session: Session = req.session as Session;
+      const token: string = retrieveAccessTokenFromSession(session);
+
+      const latestGaz1Date: string = await getLatestGaz1Date(company.companyNumber, token);
+
       return res.render(Templates.CONFIRM_COMPANY, {
         company,
+        latestGaz1Date,
         templateName: Templates.CONFIRM_COMPANY,
       });
     } catch (e) {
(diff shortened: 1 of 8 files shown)


mwestacott:
Ready for review but needs to be tested once sub-task https://companieshouse.atlassian.net/browse/OBJ-337 is completed.

src/controllers/confirm.company.controller.ts
```diff
@@ -27,12 +32,24 @@ const INELIGIBLE_PAGES = {
  * @param next
  */
 
-export const get = (req: Request, res: Response, next: NextFunction) => {
+export const get = async (req: Request, res: Response, next: NextFunction) => {
   if (req.session) {
     try {
       const company: ObjectionCompanyProfile = retrieveCompanyProfileFromObjectionSession(req.session);
+
+      const session: Session = req.session as Session;
```
SVarallo, line 40: could this be moved to the top of the get and used in the if?

mwestacott: I think I see what you mean - move it up and use in the call to retrieve the company profile? Will push a change.

SVarallo (approved):

src/controllers/confirm.company.controller.ts
```diff
@@ -27,12 +32,25 @@ const INELIGIBLE_PAGES = {
...
-      const company: ObjectionCompanyProfile = retrieveCompanyProfileFromObjectionSession(req.session);
+      const company: ObjectionCompanyProfile = retrieveCompanyProfileFromObjectionSession(session);
+
+      const token: string = retrieveAccessTokenFromSession(session);
+
+      let latestGaz1Date: string;
+      if (await getCompanyEligibility(company.companyNumber, token)) {
+        latestGaz1Date = await getLatestGaz1Date(company.companyNumber, token);
+      } else {
+        latestGaz1Date = "No notice in The Gazette";
```
markpit, line 48: i couldn't see a test to cover the 'else' line 48 scenario

markpit: i can see it now, it's been added to an existing one

mwestacott: Yes.

src/controllers/confirm.company.controller.ts
```diff
@@ -69,3 +87,20 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
 const getIneligiblePage = (apiError: ApiError): string => {
   return INELIGIBLE_PAGES[apiError.data.status];
 };
+
+const getLatestGaz1Date = async (companyNumber: string, token: string): Promise<string> => {
```
markpit, line 91: would this function be better placed in the company.filing.history.service? you could have it return the latest GAZ1 CompanyFilingHistory and then the controller could pick out the date from it. It should make it more re-usable and keep the controller a bit more 'dumb'.

mwestacott: Yep. Makes sense to me. I'm happy to action this change. @ICoxCH do you agree?

mwestacott: Changed. Function remains to extract and format the date, but other logic moved to the service.

src/services/company.filing.history.service.ts
```diff
@@ -22,4 +35,8 @@ export const getCompanyFilingHistory = async (companyNumber: string, category: s
   logger.debug("Data from company filing history SDK call " + inspect(sdkResponse));
 
   return sdkResponse.resource as CompanyFilingHistory;
+};
+
+function isGaz1(element: FilingHistoryItem) {
```
markpit, line 40: can you make this an arrow function to follow suite const ... => etc...

markpit (approved):

PR #127, opened 2020-10-08
OBJ-324 added api call to get company eligibility

SVarallo:
Adding a method to the sdk/service and objection.service classes to get the company eligibility boolean from the API.

### DIFF
--- a/src/modules/sdk/objections/service.ts
+++ b/src/modules/sdk/objections/service.ts
@@ -39,8 +39,8 @@ const OBJECTIONS_API_SINGLE_ATTACHMENT_URL =
  * @returns {boolean} the eligibility for the given company
  * @throws {ApiError}
  */
-export const getEligibility = async (companyNumber: string, token: string): Promise<boolean> => {
-  logger.debug(`getting eligibility for company number ${companyNumber}`);
+export const getCompanyEligibility = async (companyNumber: string, token: string): Promise<boolean> => {
+  logger.debug(`getting eligibility for company number ${companyNumber} from the SDK`);
 
   const axiosConfig: AxiosRequestConfig = getBaseAxiosRequestConfig(
     HTTP_GET, OBJECTIONS_ELIGIBILITY_URL(companyNumber), token);
--- a/src/modules/sdk/objections/service.ts
+++ b/src/modules/sdk/objections/service.ts
@@ -14,6 +14,9 @@ import {
   ObjectionPatch,
 } from "./types";
 
+const OBJECTIONS_ELIGIBILITY_URL = (companyNumber: string): string =>
+  `${INTERNAL_API_URL}/company/${companyNumber}/strike-off-objections/eligibility`;
+
 const OBJECTIONS_API_URL = (companyNumber: string): string =>
   `${INTERNAL_API_URL}/company/${companyNumber}/strike-off-objections`;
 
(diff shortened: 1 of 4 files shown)


test/modules/sdk/objections/service.spec.unit.ts
```diff
@@ -57,6 +57,24 @@ describe("objections SDK service unit tests", () => {
     mockGetBaseAxiosRequestConfig.mockClear();
   });
 
+  it("should return true when getting company eligibility", async () => {
```
mattch1, line 60: Could we get a test to cover the false value as well?

src/modules/sdk/objections/service.ts
```diff
@@ -27,6 +30,23 @@ const OBJECTIONS_API_SINGLE_ATTACHMENT_URL =
...
+/**
+ * Get the eligibility for the given company.
+ *
+ * @param {string} companyNumber the company number
+ * @param {string} token the bearer security token to use to call the api
+ *
+ * @returns {boolean} the eligibility for the given company
+ * @throws {ApiError}
+ */
+export const getEligibility = async (companyNumber: string, token: string): Promise<boolean> => {
```
mwestacott, line 42: Should method name be 'eligible' perhaps, or 'isEligible'? Not sure of the Node standard on this.

I see that in PTF and Extensions we had an 'exists' method that returned a boolean...

src/modules/sdk/objections/service.ts
```diff
@@ -27,6 +30,23 @@ const OBJECTIONS_API_SINGLE_ATTACHMENT_URL =
...
+  logger.debug(`getting eligibility for company number ${companyNumber}`);
```
mwestacott, line 43: I would change this log message slightly, just so that it differs from the one in the service (which I see is exactly the same). Maybe add 'from the SDK' onto the end of it?

test/modules/sdk/objections/service.spec.unit.ts
SVarallo: We could do, but I figured since you're forcing the result anyway so long as it gets the right thing back that would be enough.

src/modules/sdk/objections/service.ts
SVarallo: Yeah? Wasn't sure on the naming convention and figured I'd copy the other methods in the class

mattch1:
I cannot see what is making a call to the objection service - if nothing is we can leave it as it is, otherwise we would need to feature flag this.

mwestacott:
> I cannot see what is making a call to the objection service - if nothing is we can leave it as it is, otherwise we would need to feature flag this.

@mattch1 PR for OBJ-321 will make this call. I'm just waiting on this PR to be merged.

src/modules/sdk/objections/service.ts
markpit: how about getCompanyEligibility ?

test/modules/sdk/objections/service.spec.unit.ts
markpit: it should be ok with just true (or false) as there's no different lines of code to execute if it is one value or the other. We're testing that the value returned in the data gets converted to a boolean and returned so that should be enough.

markpit: actually, you could rename the test to be more generic, "should return boolean when getting company eligibility" I guess with "true" in the title it makes it look like it will only return true, or that there should also be a "false" version.

test/services/objection.service.spec.unit.ts
```diff
@@ -102,6 +103,17 @@ describe("objections API service unit tests", () => {
     };
   });
 
+  it("returns true when getting company eligibility", async () => {
```
markpit, line 106: change 'true' in description to 'boolean'

markpit:
the pull request could do with a description as well

markpit (approved):

mwestacott (approved):

PR #130, opened 2020-10-14
OBJ-313 fixing axe violation in accessibility statement

SVarallo:


mwestacott:
JIRA ticket id in branch name looks wrong.

PR #132, opened 2020-10-14
OBJ-313 fixing axe violation in accessibility statement

SVarallo:


ICoxCH (approved):

mwestacott:
Just wondering, I can’t see where you’ve addressed “List element has direct children that are not allowed inside <li> elements.” in your PR.

SVarallo:
> Just wondering, I can’t see where you’ve addressed “List element has direct children that are not allowed inside < li > elements.” in your PR.

Yeah, that confused me a bit as well, I think it was referring to that contact us part? Either way, I ran an axe analysis on the new page and nothing was flagged.

mwestacott (approved):

PR #133, opened 2020-10-14
OBJ-329 amending objections landing page

SVarallo:
added "What is The Gazette?" details element to the landing page with links.

### DIFF
--- a/views/index.html
+++ b/views/index.html
@@ -45,6 +45,14 @@
         <li>to upload evidence to support the objection, for example invoices showing the company is still trading or owes a debt</li>
       </ul>
 
+      {{ govukDetails({
+        summaryText: "What is the Gazette?",
+        html: '<p> Official notices about insolvency or companies being wound up are published in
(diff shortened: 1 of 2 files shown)


views/index.html
```diff
@@ -45,6 +45,14 @@ <h2 class="govuk-heading-m">Before you start</h2>
         <li>to upload evidence to support the objection, for example invoices showing the company is still trading or owes a debt</li>
       </ul>
 
+      {{ govukDetails({
+        summaryText: "What is The Gazette?",
```
markpit, line 49: how about a unit test to 'GET' the page and check it contains the text - What is The Gazette - just to cover it off if it got accidentally deleted in future.

mwestacott (approved):
Not convinced that we need to add unit tests to check every bit of text on the screen (unless its display/presence/content is controlled by some bit of logic), also, testers have tests that do this (so some duplication of effort) but no harm in doing this so have approved.

PR #136, opened 2020-10-15
OBJ-333 Amendments to the 'Check your objection application' screen

mwestacott:


views/check-your-answers.html
```diff
@@ -54,7 +47,7 @@ <h2 class="govuk-heading-m">
           },
           {
             key: {
-              text: "Full name"
+              text: "Full name or organisation name"
             },
             value: {
               text: objection.created_by.full_name
```
mwestacott, line 53: I did think about changing the ‘full_name’ field name to something else now that organisation name is mentioned in the labels (just ‘name’ perhaps?) but decided against this - it appears in a bunch of places in the web app but then also travels into the API and features in a lot of places there too (in the Swagger spec also).

ICoxCH (approved):

mattch1 (approved):

PR #137, opened 2020-10-16
Feature/obj 250 remove non printable characters

mattch1:
Copied from extensions but added regex to ensure that we have single spaces between words and added testing for this.


src/utils/string.formatter.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * String formatting functions
+ */
+const REGEX_NON_PRINTABLE_CHARS: RegExp = /[^ -\xFF€]+/g;
+const REGEX_MULTIPLE_SPACES: RegExp = /\s\s+/g;
```
ICoxCH, line 5: Does this need to be a regex? would "  " not do?
Also what happens if there are 3 or more spaces?

src/utils/string.formatter.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * String formatting functions
+ */
+const REGEX_NON_PRINTABLE_CHARS: RegExp = /[^ -\xFF€]+/g;
```
ICoxCH, line 4: would the regex `/[^\x00-\x7F]/g` be better? (match all non ASCII characters)
This regex seems to be match one or more, SPACE to ÿ, and € why this range specifically?

src/utils/string.formatter.ts
mattch1: Yes we should use regex for this reason - for your second querstion have a look at the last unit test I added.

src/utils/string.formatter.ts
mattch1: I'm not sure we want to exclude all non ascii characters and I think we want to start at 32 DEC rather than 0 so that we don't have \n etc.

ICoxCH: is there any guidance on what is a "Non printable character"?

mattch1: Good question

ICoxCH: I've found this? https://en.wikipedia.org/wiki/Non-printing_character_in_word_processors

mattch1: Looked here https://web.itu.edu.tr/sgunduz/courses/mikroisl/ascii.html and foudn that non printable is 0-32 and printable is the extensed ascii set 32-255.

ICoxCH: Looking at that then, and the wikipedia article, should the regex match 0-31 (inclusive), so when calling str.replace it removes those characters?

mattch1: From what I remember from the extension work I think we want to included extended ascii - i'll see if I can find the original ticket for that.

ICoxCH: May be worth flagging the story and asking for more detailed requirements? It shouldn't really be down to us to decide which characters are or aren't "printable"

mattch1: To address your question are we suggesting an explict positive reference to 0-31 rather than anything that is not 32-255?

ICoxCH: If the definition of a non printable character is 0-31. Then yes we should only match and remove 0-31.
As Not 32-255 would remove characters over 255, which if definition for non printable is 0-31, then over 255 is still printable.
We need a more concrete definition of what the story considers to be a non printable character

mattch1: I think the current definition is specifically to limit this to just the ascii and extended ascii set - but it would be worth verifying this.

mattch1: Just a final note - I'm guessing the reason we used this for extensions is so that we don't leave ourselves too exposed to any potential complications that a  >255 might bring.

src/utils/string.formatter.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * String formatting functions
+ */
+const REGEX_NON_PRINTABLE_CHARS: RegExp =
```
ICoxCH, line 4: could this be renamed to REGEX_MATCH_NON_PRINTABLE_CHARS ? to make it clearer what the regex does

test/utils/string.formatter.spec.unit.ts
```diff
@@ -0,0 +1,55 @@
+import { removeNonPrintableChars } from "../../src/utils/string.formatter";
+
+
+describe("String formatter tests", () => {
+  it("should not remove formatting chars", () => {
+    const dodgyString: string = "hello.\r\nThis string is not dodgy!";
```
ICoxCH, line 6: perhaps rename this (for all the tests) to preFormattedString

test/utils/string.formatter.spec.unit.ts
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  it("should return original string if it is blank", () => {
+    const blankString: string = " ";
+    const formattedString: string = removeNonPrintableChars(blankString);
+
+    expect(formattedString).toEqual(blankString);
+  });
+
+  //"Hylô amgáu düwch mẁg"
```
ICoxCH, line 54: is this comment intended?

test/utils/string.formatter.spec.unit.ts
mattch1: I like "dodgyString"

src/utils/string.formatter.ts
mattch1: I'd rather keep it short

test/utils/string.formatter.spec.unit.ts
mattch1: No, removed this.

src/utils/string.formatter.ts
ICoxCH: the name leads me to believe the regex is a list of non printable characters, which is the opposite of what it is. A name that makes it clear it is matching non printable characters would be ideal

test/utils/string.formatter.spec.unit.ts
ICoxCH: but this example, it is not a dodgy string

src/utils/string.formatter.ts
mattch1: Ok renamed.

test/utils/string.formatter.spec.unit.ts
mattch1: Renamed these.

src/utils/string.formatter.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * String formatting functions
+ */
+const REGEX_MATCH_NON_PRINTABLE_CHARS: RegExp =
+    /[^ÀÁÂÄÈÉÊËÌÍÎÏÒÓÔÖÙÚÛÜẀẂŴẄỲÝŶŸàáâäèéêëìíîïòóôöùúûüẁẃŵẅỳýŷÿABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789&@£$€¥.,:;–\-'''()\[\]{}<>!""»«"\?\/\\*=#%\+ \r\n]/g;
```
ICoxCH, line 5: I think sonar is complaining because {} aren't escaped.
Try \{\} and put a \ in front of +

mattch1: I cannot see this on Sonar.

mattch1: Lint complains when you do escape then  - what advice does Sonar give.

ICoxCH: http://code-analysis.platform.aws.chdev.org:9000/dashboard?id=uk.gov.companieshouse%3Astrike-off-objections-web&pullRequest=137

This is the sonar report, essentially needs less +*{ characters

ICoxCH: May also be worth simplifying A-Za-z0-9 for easier readability.
You seem to have char code 39 (') 3 times, you only need it once, same for char code 34 (")

mattch1: Simplifyed. The repeated ' is as miscopy it should be ` changed this.

ICoxCH:
Unfortunately sonar is still not happy, I would suggest creating a set of accepted characters, and iterate over the input, and if the character is not in accepted set, removing it

src/controllers/enter.information.controller.ts
```diff
@@ -29,7 +30,7 @@ export const post = async (req: Request, res: Response, next: NextFunction) => {
   const company: ObjectionCompanyProfile = retrieveCompanyProfileFromObjectionSession(session);
   const objectionId: string = retrieveFromObjectionSession(session, SESSION_OBJECTION_ID);
 
-  const reason: string = req.body.information;
+  const reason: string = removeNonPrintableChars(req.body.information);
```
mwestacott, line 33: In theory might need to apply this to the only other text field where the user enters data - the 'full name / organisation name' field - but not mentioned in story so this PR is fine.

src/utils/string.formatter.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * String formatting functions
+ */
+const PRINTABLE_CHARS: string =
```
ICoxCH, line 4: Recommend using a set: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Set

As we can then use if PRINTABLE_CHARS.contains(char); which is more efficient

mattch1: How much more effiicent way more efficient or just little bit more?

ICoxCH: Assuming sets in typescript work the same as they do in Java.
You can tell if an item is in a set with a single calculation.
Whereas to tell if an item is in a list (or in this case string) it needs to do a calculation for each index of the list. (In this case 158) so by using a set we save 158 calculations for each character in the input string.

mattch1: I'm aware of this but the trouble is we are offered this set but no nice way of turning a string into a set .

ICoxCH: const PRINTABLE_CHARS: Set = new Set(['A', 'B', 'C']); 
This would initialise a new set with values. Then:

    for (let index = 0; index < inputStr.length; index++) {
      if (PRINTABLE_CHARS.has(inputStr.charAt(index))) {
        inputStr = inputStr.replace(inputStr.charAt(index), " ");
      }
    }

mattch1: That's what I was afraid of.

ICoxCH: We 'could' use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from
But personally I would be against this as it makes the code slightly less readable. 
Happy to get more opinions on if using this is allowed

markpit: you could loop through the string using [...text].forEach(). see https://stackoverflow.com/a/22410812 . But i guess it would still be computationally expensive.

mattch1: Found a way to do it.

src/utils/string.formatter.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * String formatting functions
+ */
+const PRINTABLE_CHARS: Set<String> =
```
markpit, line 4: shouldn't these be chars?

markpit: ignore me, charAt returns a string!

mattch1: Unfortunately, this aint java

mattch1: The set requires a generic type you see.

src/utils/string.formatter.ts
markpit: how about this

const PRINTABLE_CHARS: string = "etc...."
const mySet: Set<string> = new Set<string>([...PRINTABLE_CHARS]);

src/utils/string.formatter.ts
ICoxCH: One last tiny thing from me, then i'll approve I swear.
Sonar has produced a smell, we should use primitive string here, instead of the wrapper type String

http://code-analysis.platform.aws.chdev.org:9000/project/issues?id=uk.gov.companieshouse%3Astrike-off-objections-web&open=AXVqfUZFbKZCo3W_hI8L&pullRequest=137&resolved=false&sinceLeakPeriod=true&types=CODE_SMELL

src/utils/string.formatter.ts
mattch1: Already sorted this - I don't know whetehr there isn't some operation going on under the hood for this - I think the set has been chosen because we want to start with a set.

src/utils/string.formatter.ts
```diff
@@ -0,0 +1,16 @@
...
+    new Set(["À", "Á", "Â", "Ä", "È", "É", "Ê", "Ë", "Ì", "Í", "Î", "Ï", "Ò", "Ó", "Ô", "Ö", "Ù", "Ú", "Û", "Ü", "Ẁ", "Ẃ", "Ŵ", "Ẅ", "Ỳ", "Ý", "Ŷ", "Ÿ", "à", "á", "â", "ä", "è", "é", "ê", "ë", "ì", "í", "î", "ï", "ò", "ó", "ô", "ö", "ù", "ú", "û", "ü", "ẁ", "ẃ", "ŵ", "ẅ", "ỳ", "ý", "ŷ", "ÿ", "A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z", "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "&", "@", "£", "$", "€", "¥", ".", ", ", ":", ";", "–", "-", "‘", "’", "'", "(", ")", "[", "]", "{", "}", "<", ">", "!", "“", "”", "»", "«", "\"", "?", "/", "\\", "*", "=", "#", "%", "+", " ", "\r", "\n"]);
+export const removeNonPrintableChars = (inputStr: string): string => {
+  if (inputStr) {
+    for (let index = 0; index < inputStr.length; index++) {
+      if (!PRINTABLE_CHARS.has(inputStr.charAt(index))) {
```
markpit, line 9: you have 2 calls to inputStr.charAt(index), you can do this just once and store in a var.

src/utils/string.formatter.ts
mattch1: Right using "string"

src/utils/string.formatter.ts
mattch1: Single call

src/utils/string.formatter.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * String formatting functions
+ */
+const PRINTABLE_CHARS: Set<string> =
+    new Set(["À", "Á", "Â", "Ä", "È", "É", "Ê", "Ë", "Ì", "Í", "Î", "Ï", "Ò", "Ó", "Ô", "Ö", "Ù", "Ú", "Û", "Ü", "Ẁ", "Ẃ", "Ŵ", "Ẅ", "Ỳ", "Ý", "Ŷ", "Ÿ", "à", "á", "â", "ä", "è", "é", "ê", "ë", "ì", "í", "î", "ï", "ò", "ó", "ô", "ö", "ù", "ú", "û", "ü", "ẁ", "ẃ", "ŵ", "ẅ", "ỳ", "ý", "ŷ", "ÿ", "A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z", "0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "&", "@", "£", "$", "€", "¥", ".", ", ", ":", ";", "–", "-", "‘", "’", "'", "(", ")", "[", "]", "{", "}", "<", ">", "!", "“", "”", "»", "«", "\"", "?", "/", "\\", "*", "=", "#", "%", "+", " ", "\r", "\n"]);
+export const removeNonPrintableChars = (inputStr: string): string => {
```
markpit, line 6: should have a new line between string and start of function

markpit (approved):

ICoxCH (approved):

PR #141, opened 2020-10-20
OBJ-111 add page not found screen for 404 errors

mattch1:
Added new page template, test and error function for 404 errors.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -3,7 +3,7 @@ import express from "express";
 import * as nunjucks from "nunjucks";
 import * as path from "path";
 import { APP_NAME } from "./constants";
-import { errorHandler } from "./controllers/error.controller";
+import  errorHandlers from "./controllers/error.controller";
 import { authenticationMiddleware } from "./middleware/authentication.middleware";
 import { objectionSessionMiddleware } from "./middleware/objection.session.middleware";
 import { serviceAvailabilityMiddleware } from "./middleware/service.availability.middleware";
@@ -45,7 +45,7 @@ app.use(`${pageURLs.STRIKE_OFF_OBJECTIONS}/*`, authenticationMiddleware);
 app.use(`${pageURLs.STRIKE_OFF_OBJECTIONS}/*(?<!download)$`, objectionSessionMiddleware);
 // apply our default router to /
 app.use(pageURLs.STRIKE_OFF_OBJECTIONS, router);
-app.use(errorHandler);
+app.use(...errorHandlers);
 
 logger.info(`************** ${APP_NAME} has started **************`);
 
--- a/src/model/template.paths.ts
+++ b/src/model/template.paths.ts
@@ -18,5 +18,5 @@ export enum Templates {
   ERROR = "error",
   FILE_ERROR = "file-error",
   ACCESSIBILITY_STATEMENT = "accessibility-statement",
-  ERROR_404 = "page-not-found.html"
+  ERROR_404 = "page-not-found"
 }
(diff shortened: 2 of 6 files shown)


src/app.ts
```diff
@@ -3,7 +3,7 @@ import express from "express";
 import * as nunjucks from "nunjucks";
 import * as path from "path";
 import { APP_NAME } from "./constants";
-import { errorHandler } from "./controllers/error.controller";
+import  errorHandlers from "./controllers/error.controller";
```
ICoxCH, line 6: Not sure functionally it'll make a difference, but all other imports are surrounded by curly braces, we should porobably be consistent

src/model/template.paths.ts
```diff
@@ -18,4 +18,5 @@ export enum Templates {
   ERROR = "error",
   FILE_ERROR = "file-error",
   ACCESSIBILITY_STATEMENT = "accessibility-statement",
+  ERROR_404 = "page-not-found.html"
```
markpit, line 21: the other names don't seem to have '.html' on them, will it work without?

test/controller/error.controller.spec.unit.ts
```diff
@@ -1,23 +1,62 @@
...
+mockAuthenticationMiddleware.mockImplementation((req: Request, res: Response, next: NextFunction) => next());
 
+const DODGY_URL = STRIKE_OFF_OBJECTIONS  + "/company-numberhh";
 const ERROR_500 = "Sorry, there is a problem with the service";
 
 describe("error controller", () => {
 
+  it("should render the 404 template if a page is not found", async () => {
+    mockSessionMiddleware.mockReset();
+    mockSessionMiddleware.mockImplementation((req: Request, res: Response, next: NextFunction) => {
```
markpit, line 30: i would use mockImplementationOnce instead of mockImplementation as we are defining them inside this test, we only need the implementation to be applied just for this test. If we don't, the implementation will apply to the tests that run after (unless they explicitly override the implementation)

test/controller/error.controller.spec.unit.ts
```diff
@@ -1,23 +1,62 @@
...
+      .set("referer", "/")
+      .set("Cookie", [`${COOKIE_NAME}=123`]);
+
+    expect(resp.status).toEqual(404);
+    expect(resp.text).toContain(ErrorMessages.ERROR_404);
+  });
+
   it("should render the error template", async () => {
+    mockSessionMiddleware.mockReset();
+    mockSessionMiddleware.mockImplementation((req: Request, res: Response, next: NextFunction) => {
```
markpit, line 57: same here for the Once method

src/app.ts
markpit: just tried it in mine, ide doesn't like it with brackets, I think because it is the default export.

ICoxCH: fair enough ignore this comment then

test/controller/error.controller.spec.unit.ts
mattch1: Done.

test/controller/error.controller.spec.unit.ts
mattch1: Ok

src/model/template.paths.ts
mattch1: Don't need the html part - removed.

test/controller/error.controller.spec.unit.ts
```diff
@@ -1,23 +1,62 @@
...
+  it("should render the 404 template if a page is not found", async () => {
+    mockSessionMiddleware.mockReset();
+    mockSessionMiddleware.mockImplementationOnce((req: Request, res: Response, next: NextFunction) => {
+      req.session = {
+        data: {},
+      } as Session;
+      return next();
+    });
+
+    mockObjectionSessionMiddleware.mockImplementation((req: Request, res: Response, next: NextFunction) => {
```
markpit, line 37: could you Once this one as well?

markpit (approved):

ICoxCH (approved):

PR #156, opened 2020-10-28
OBJ-384 bugfix for printable chars comma

mattch1:
Comma should have been "," not ", " added this to the unit test and checked that other charcaters do not have trailing spaces.


ICoxCH (approved):

test/utils/string.formatter.spec.unit.ts
```diff
@@ -31,10 +31,10 @@ describe("String formatter tests", () => {
   });
 
   it("should not remove printable chars", () => {
```
ICoxCH, line 33: I'd suggest adding a unit test where the preFormattedString is a string of the entire set of acceptable characters, and then check the length of the returned string === length of the preFormattedString.

markpit: i'd agree, we should test all the allowable characters are actually allowed

mattch1: Added the new test.

ICoxCH (approved):

test/utils/string.formatter.spec.unit.ts
```diff
@@ -30,11 +29,19 @@ describe("String formatter tests", () => {
...
     const formattedString: string = removeNonPrintableChars(preFormattedString);
 
-    expect(formattedString).toEqual("hello. This string is not dodgy !@£$%&*{}()€#");
+    expect(formattedString).toEqual("hello. This string is not dodgy, !@£$%&*{}()€#");
+  });
+
+  it("should not remove any printable chars when entire set is provided", () => {
+    const preFormattedString: string = "ÀÁÂÄÈÉÊËÌÍÎÏÒÓÔÖÙÚÛÜẀẂŴẄỲÝŶŸàáâäèéêëìíîïòóôöùúûüẁẃŵẅỳýŷÿABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789&@£$€¥.,:;–-‘’'()[]{}<>!“”»«\"?/\\*=#%+ \r\n";
+    const formattedString: string = removeNonPrintableChars(preFormattedString);
+    const test: boolean = preFormattedString === formattedString;
```
markpit, line 42: looks like you're testing the same thing twice here. if toEqual isn't enough you can use toStrictEqual

mattch1: They are different but will change.

markpit (approved):

PR #164, opened 2020-11-03
OBJ-359 Adding mocks for index controller test and service availibili…

markpit:
…ty test


test/middleware/service.availability.middleware.spec.unit.ts
```diff
@@ -1,9 +1,36 @@
 jest.mock("ioredis");
+jest.mock("../../src/middleware/authentication.middleware");
```
markpit, line 2: now the landing page uses the middleware, the test needs the middleware session stuff mocked to pass

SVarallo (approved):

PR #165, opened 2020-11-03
OBJ-359 removing previous session data present after signing out and in

SVarallo:
Added a controller for the index page with dedicated get and push methods instead of just a render. In the push method it will now check there is a session present, and if so, will delete any extra data held in that session.


src/controllers/index.controller.ts
```diff
@@ -0,0 +1,23 @@
...
+  return res.render(Templates.INDEX, {
+    templateName: Templates.INDEX,
+  });
+};
+
+export const post = (req: Request, res: Response) => {
+  const session: Session | undefined = req.session as Session;
+  if (session) {
+    logger.debug("Deleting existing session data from " + OBJECTIONS_SESSION_NAME);
+    logger.debug(OBJECTIONS_SESSION_NAME);
```
markpit, line 18: logger.debug(OBJECTIONS_SESSION_NAME);
looks like it is redundant now

ICoxCH, line 18: Think some left over debugging logs from testing are here. I would say we only need the "Deleting existing session data from..." log statement

markpit (approved):

PR #166, opened 2020-11-04
OBJ-359 reverted start now button id back to start-now from submit

SVarallo:


ICoxCH (approved):
Let the Sonar checks come inbefore merging. But i'm confident will be fine

PR #168, opened 2020-11-06
Adding missing links field to company profile test

markpit:
company.profile.spec.unit.ts was showing a warning that field 'links' was missing from the company profile object so this is to add it.


test/services/company.profile.service.spec.unit.ts
```diff
@@ -63,6 +63,7 @@ const dummySDKResponse: Resource<CompanyProfile> = {
     hasCharges: false,
     hasInsolvencyHistory: false,
     jurisdiction: "england-wales",
+    links: {},
```
mattch1, line 66: Are we including any test data in here?

markpit: what would we test?

mattch1 (approved):

ICoxCH (approved):

PR #172, opened 2020-11-17
OBJ-248 create objection logging error 400 fix

markpit:
We were previously returning a 400 from the api when creating an objection request for an ineligible company which would result in Axios throwing an error which was appearing in the logs as an error and caught in the confirm.company controller and handled from there. this was misleading as the request had been created and no real error had occurred.
  This change is in line with a change to the api (v1.1.0) to change the http response status from the api when creating an objection to be 201, even when the company is ineligible but the response will now contain both the id and status. These changes are to capture those values and to decide what page to show based on the status.

Also improved the axios logging to reduce number of 'circular references' from axios api calls which were preventing debug logging from showing properly in the logs.

depends on api change - https://github.com/companieshouse/strike-off-objections-api/pull/106


src/modules/sdk/objections/axios.client.ts
```diff
@@ -39,7 +39,11 @@ export const makeAPICall = async (config: AxiosRequestConfig): Promise<AxiosResp
     logger.debug(`Calling ${config.method} ${config.url}`);
     const axiosResponse: AxiosResponse = await axios.request<any>(config);
     try {
-      logger.debug(`axios response from api call : ${JSON.stringify(axiosResponse, null, 2)}`);
+      const loggableAxiosResponse = {
```
markpit, line 42: thought i'd improve the logging from Axios a bit. we're only really interested in the status and data. The issues of 'circular references' were mostly references in the root level of the response object and not in 'data'. So logging the data and status allows us to avoid most of the issues. Still have an issue when logging a download but not worried about that as the 'data' is just bytes. The file header info gets logged elsewhere.

markpit: and added a test to make sure we are logging these values.

src/modules/sdk/objections/service.ts
```diff
@@ -57,13 +57,18 @@ export const getCompanyEligibility = async (companyNumber: string, token: string
  * @returns {string} the id of the newly created objection
  * @throws {ApiError}
  */
-export const createNewObjection = async (companyNumber: string, token: string, createWithData: ObjectionCreate): Promise<string> => {
+export const createNewObjection = async (companyNumber: string, token: string, createWithData: ObjectionCreate): Promise<ObjectionCreatedResponse> => {
```
markpit, line 60: api change means it returns id and status so I've created an ObjectionCreatedResponse type to capture these. It used to just return the id as a string.

ICoxCH (approved):

PR #178, opened 2020-12-10
obj-447 gaz2 requested or issued show notice expired message

mattch1:
Confirm this is the correct company screen will display ‘Notice expired’ if either:
EligibilityStatus.INELIGIBLE_GAZ2_REQUESTED 
EligibilityStatus.INELIGIBLE_COMPANY_STRUCK_OFF

are returned.

### DIFF
--- a/src/controllers/confirm.company.controller.ts
+++ b/src/controllers/confirm.company.controller.ts
@@ -4,7 +4,7 @@ import ObjectionCompanyProfile from "model/objection.company.profile";
 import { SESSION_OBJECTION_ID } from "../constants";
 import { OBJECTIONS_ENTER_INFORMATION, OBJECTIONS_NO_STRIKE_OFF, OBJECTIONS_NOTICE_EXPIRED } from "../model/page.urls";
 import { Templates } from "../model/template.paths";
-import { CompanyEligibility, EligibilityStatus, ObjectionCreate, ObjectionStatus } from "../modules/sdk/objections";
+import { EligibilityStatus, ObjectionCreate, ObjectionStatus } from "../modules/sdk/objections";
 import { createNewObjection, getCompanyEligibility } from "../services/objection.service";
 import {
   addToObjectionSession,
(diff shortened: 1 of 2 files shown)


test/controller/confirm.company.controller.spec.unit.ts
```diff
@@ -256,13 +256,14 @@ describe("confirm company tests", () => {
...
 
-  it("should correctly display the latest GAZ1 date from the filing history if company in gaz1, with gaz2 requested", async () => {
+  it("should correctly display the notice expired message from the filing history if company in gaz1, with gaz2 requested", async () => {
 
     const mockValidAccessToken = retrieveAccessTokenFromSession as jest.Mock;
     mockValidAccessToken.mockReset().mockImplementation(() => ACCESS_TOKEN);
 
     mockGetObjectionSessionValue.mockReset().mockImplementation(() => dummyCompanyProfile);
 
+    mockGetCompanyEligibility.mockReset();
```
markpit, line 266: you shouldn't need to add the mock reset as it's happening also in the next line, so it is being Reset twice.

test/controller/confirm.company.controller.spec.unit.ts
```diff
@@ -275,7 +276,28 @@ describe("confirm company tests", () => {
...
+  });
+
+  it("should correctly display the notice expired message from the filing history if company is INELIGIBLE_COMPANY_STRUCK_OFF", async () => {
+
+    const mockValidAccessToken = retrieveAccessTokenFromSession as jest.Mock;
+    mockValidAccessToken.mockReset().mockImplementation(() => ACCESS_TOKEN);
+
+    mockGetObjectionSessionValue.mockReset().mockImplementation(() => dummyCompanyProfile);
+
+    mockGetCompanyEligibility.mockReset();
```
markpit, line 289: same here for the duplicate reset

src/controllers/confirm.company.controller.ts
```diff
@@ -42,9 +42,12 @@ export const get = async (req: Request, res: Response, next: NextFunction) => {
 
       let latestGaz1Date: string;
       const companyEligibility: CompanyEligibility = await getCompanyEligibility(company.companyNumber, token);
-      if (companyEligibility.eligibility_status === EligibilityStatus.ELIGIBLE ||
-        companyEligibility.eligibility_status === EligibilityStatus.INELIGIBLE_GAZ2_REQUESTED) {
+
+      if (companyEligibility.eligibility_status === EligibilityStatus.ELIGIBLE) {
```
markpit, line 46: This is a personal preference one, but nodejs will allow you to capture just the fields you want from an object when it is being assigned.
instead of 
`const companyEligibility: CompanyEligibility = await getCompanyEligibility(company.companyNumber, token);`
you can have `const {eligibility_status} = await getCompanyEligibility(company.companyNumber, token);`
then you can just use eligibility_status in the if statements. It will reduce the code but it's just a different way i guess, down to what you prefer.

markpit (approved):

src/controllers/confirm.company.controller.ts
mattch1: Simplified the code.

markpit (approved):

PR #195, opened 2021-04-08
Creating a skeleton design page for objector type selection

YiannisPapastergiou:
Creating a skeleton design page for objector type selection.
This page does not contain any functionality.
Only necessary information like route and paths added.

Resolves: [OBJ-511](https://companieshouse.atlassian.net/browse/OBJ-511)


mouhajer-ch (approved):
It looks good to me!

PR #197, opened 2021-04-15
Adding matomo events for 'Who is applying to object?' screen

toriwilliamsch:
Adding matomo events for the who is applying to object screen

Resolves: [OBJ-443](https://companieshouse.atlassian.net/browse/OBJ-443)


mouhajer-ch (approved):
It looks good!

PR #200, opened 2021-04-20
Feature/obj 438 update objecting entity name page based on objector selected

mouhajer-ch:
- Added logic that dynamically and based on the feature flag updates the object entity name page. 
- Added dynamic back link from entity name page based on feature flag on/off to point to start now page or the new page
- Added logic to save objector selected when we go back from objector-organisation page on session ExtraData object
- Added Unit tests

Resolves [OBJ-438](https://companieshouse.atlassian.net/browse/OBJ-438)

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -12,7 +12,7 @@ export const CHANGE_ANSWER_KEY = "change";
 
 
 export const FULL_NAME_FIELD = "fullName";
-export const DIVULGE_INFO_FIELD = "shareIdentity";
+export const SHARE_IDENTITY_FIELD = "shareIdentity";
 export const GENERIC_INFO = "generic";
 export const OBJECTOR_FIELDS = {
   "myself-or-company": {
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,3 +1,7 @@
+import {
+  OBJECTIONS_OBJECTOR_ORGANISATION,
+  STRIKE_OFF_OBJECTIONS } from "./model/page.urls";
+
 export const APP_NAME = "strike-off-objections-web";
 export const OBJECTIONS_SESSION_NAME = "strike_off_objections_session";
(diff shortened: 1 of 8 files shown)


src/constants.ts
```diff
@@ -1,6 +1,33 @@
...
 export const OBJECTIONS_SESSION_NAME = "strike_off_objections_session";
 export const SESSION_COMPANY_PROFILE = "objections_company_profile";
 export const SESSION_OBJECTION_CREATE = "objection_create";
 export const SESSION_OBJECTION_ID = "objection_id";
+export const SESSION_OBJECTOR = "objector";
 export const CHANGE_ANSWER_KEY = "change";
+
+
+export const FULL_NAME_FIELD = "fullName";
+export const DIVULGE_INFO_FIELD = "shareIdentity";
```
toriwilliamsch, line 15: Could this be renamed to `SHARE_IDENTITY_FIELD` please?

views/objecting-entity-name.html
```diff
@@ -25,13 +25,10 @@
       <form method="post">
         {{ govukInput({
           label: {
-            text: "What is your full name or the name of your organisation?",
+            text: textFullName,
             isPageHeading: true,
             classes: "govuk-label--xl"
```
toriwilliamsch, line 30: Would you be able to change the class to be `govuk-label--l` instead please? This will then match the prototype

toriwilliamsch:
Tested all scenarios for the feature flag, back links and dynamic text populated, all works great. Just 2 minor comments 👍

src/constants.ts
mouhajer-ch: Thanks Tori! Fix done

views/objecting-entity-name.html
mouhajer-ch: Thanks Tori! Fix done

toriwilliamsch (approved):

PR #256, opened 2022-09-16
Improve accessibility on sign-out page

daisysanson:
Resolves: [BI-11638](https://companieshouse.atlassian.net/browse/BI-11637)

- Update title to have the a page heading value that will register as h1 heading to prevent h1 missing alert in WAVE.

-  Amend typo in the signout radio id and fix href link in the error message to direct to the correct 'yes' radio attribute when a error is triggered

### DIFF
--- a/views/signout.html
+++ b/views/signout.html
@@ -31,7 +31,7 @@
             errorList: [
                 {
                 text: "Select yes if you want to sign out",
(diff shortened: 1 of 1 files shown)


adr1ancho (approved):
Looks Good To Me

PR #261, opened 2022-12-08
BI-11957/Stop new reports being created if company already selected

daisysanson:
Resolves [BI-11957](https://companieshouse.atlassian.net/browse/BI-11957)
- Add another call to addObjection and previously selected company to session into GET call of `enter information` page so that it is saved in session when the user traverses back to confirm company page.
- Add unit test to cover this call to set previously selected company

### DIFF
--- a/src/controllers/enter.information.controller.ts
+++ b/src/controllers/enter.information.controller.ts
@@ -34,7 +34,6 @@ const postEnterInformation = async (req: Request, res: Response, next: NextFunct
 
     if (!errors.isEmpty()) {
       const informationError: GovUkErrorData = createErrorData(errors, ENTER_INFORMATION);
- 
       return res.render(Templates.ENTER_INFORMATION, {
         errorList: [informationError],
         informationError,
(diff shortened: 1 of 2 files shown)


src/controllers/enter.information.controller.ts
```diff
@@ -86,8 +87,11 @@ const getObjectionFromSession = async (req: Request) => {
     throw new Error("No session present");
   }
   const objection: Objection = await getObjection(req.session);
+  const companyProfileInSession: ObjectionCompanyProfile = retrieveFromObjectionSession(req.session, SESSION_COMPANY_PROFILE);
+  if (companyProfileInSession) {
+  addToObjectionSession(req.session, PREVIOUSLY_SELECTED_COMPANY, companyProfileInSession.companyNumber);
```
adr1ancho, line 92: identation

src/controllers/enter.information.controller.ts
```diff
@@ -86,8 +87,11 @@ const getObjectionFromSession = async (req: Request) => {
     throw new Error("No session present");
   }
   const objection: Objection = await getObjection(req.session);
+  const companyProfileInSession: ObjectionCompanyProfile = retrieveFromObjectionSession(req.session, SESSION_COMPANY_PROFILE);
+  if (companyProfileInSession) {
+    addToObjectionSession(req.session, PREVIOUSLY_SELECTED_COMPANY, companyProfileInSession.companyNumber);
+  }
   if (!objection) {
```
adr1ancho, line 94: Do we still need this if check?

daisysanson: In a normal circumstance, it works either way, I guess it's just a fail safe in case it tries to add something undefined into the previously selected company value if company profile were for some reason not present.

adr1ancho (approved):
LGTM

PR #286, opened 2023-09-06
Set staging vars

sienkin:
Added vars for Staging
Removed deprecated vars
Addedf new local for `kms_alias` and updated data lookup
Corrected piwik ID for dev


tobafolaranmi-ch (approved):
LGTM

PR #296, opened 2024-01-09
code changes to fix code smells in repo

kdobija1:
JIRA Ticket: https://companieshouse.atlassian.net/browse/DEBT-1986

Remove code smells from code found by sonar

### DIFF
--- a/src/modules/sdk/objections/axios.client.ts
+++ b/src/modules/sdk/objections/axios.client.ts
@@ -54,11 +54,10 @@ export const makeAPICall = async (config: AxiosRequestConfig): Promise<AxiosResp
     logger.error(`ERROR calling API ${config.url} - ${apiErr}`);
     const axiosError = apiErr as AxiosError;
     const { response, message } = axiosError;
-     const thrownError: ApiError = {
-        data: response ? response.data : [],
-        message: message,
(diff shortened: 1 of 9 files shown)


ch-code-analysis:
CI: Security warnings found!

src/modules/sdk/objections/axios.client.ts
```diff
@@ -54,11 +54,11 @@ export const makeAPICall = async (config: AxiosRequestConfig): Promise<AxiosResp
...
-        message: message,
-        status: response ? response.status : STATUS_NO_RESPONSE,
-    };
-    throw thrownError
+    const thrownError: ApiError = {
+      data: response ? response.data : [],
+      message: message,
+      status: response ? response.status : STATUS_NO_RESPONSE,
+  };
+  throw thrownError
```
mreis-ch, line 62: The code smell says to immediately throw the expression. Was there a reason it doesn't?

kdobija1: I changed it but that just generated a different code smell over you throwing an expression instead of an error so I ended up leaving it

mreis-ch: Ok, fair enough 👍

mreis-ch (approved):

PR #333, opened 2026-08-06
Add EnsureSessionCookiePresentMiddleware to fix empty csrf issue

burhanb99:
The intermittent issue is still there.

Steps to re-produce: 
1) Delete all browsing cookies
2) Go to https://cidev.aws.chdev.org/strike-off-objections
3) Click on Make an objection

the _csrf token in the payload is blank which is causing this issue.

Fix is to use the EnsureSessionCookiePresentMiddleware as mentioned in the 
https://github.com/companieshouse/web-security-node
Readme

This PR will be accompanied by a ecs-service-configs-dev change to set COOKIE_SECURE_FLAG to 1

PR #334, opened 2026-08-06
Add dedicated healthcheck path to escape all middlewares

burhanb99:
When the ensureSessionCookiePresentMiddleware kicks in
https://github.com/companieshouse/strike-off-objections-web/pull/333

it doesn't return a 200 for /strike-off-objections, it returns 302 and then redirects back to the same url which returns 200.

This causes the healthcheck to fail cause it doesn't see a 200 and the build fails.

This pr is to fix that behaviour and add a dedicated healthcheck path before any middleware kicks in.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -52,6 +52,9 @@ app.set("views", path.join(__dirname, "views"));
 app.set("view engine", "html");
 app.use(express.static(path.join(__dirname, "public")));
 
+app.get(`${pageURLs.STRIKE_OFF_OBJECTIONS}/healthcheck`, (_req, res) => {
+  res.set("Cache-Control", "no-store").sendStatus(200);
+});
 app.use(serviceAvailabilityMiddleware);
 app.use(cookieParser());
 app.use(`${pageURLs.STRIKE_OFF_OBJECTIONS}*`, sessionMiddleware);
--- a/test/middleware/service.availability.middleware.spec.unit.ts
+++ b/test/middleware/service.availability.middleware.spec.unit.ts
@@ -83,4 +83,13 @@ describe("Availability tests", () => {
     expect(response.text).toMatch(/Sorry, the service is unavailable/);
   });
 
+  it("should return 200 for healthcheck regardless of service availability flag", async () => {
+  process.env.SHOW_SERVICE_OFFLINE_PAGE = "true";
+
+  const response = await request(app)
+    .get("/strike-off-objections/healthcheck");
+
+  expect(response.status).toEqual(200);
+});
+
 });


ch-code-analysis:
CI: No security warnings found

src/app.ts
```diff
@@ -52,6 +52,9 @@ app.set("views", path.join(__dirname, "views"));
 app.set("view engine", "html");
 app.use(express.static(path.join(__dirname, "public")));
 
+app.get(`${pageURLs.STRIKE_OFF_OBJECTIONS}/healthcheck`, (_req, res) => {
+  res.set("Cache-Control", "no-store").sendStatus(200);
+});
```
bcullerton, line 57: Doesn't this still get caught by the session middleware checks. The middleware catches ${pageURLs.STRIKE_OFF_OBJECTIONS}* so the healthcheck would be subject to those checks as things stand?

src/app.ts
bcullerton, line 61: See https://github.com/companieshouse/strike-off-objections-web/pull/334/changes#r3734276983

ch-code-analysis:
CI: No security warnings found

src/app.ts
burhanb99: Hi no, the healthcheck route is registered with app.get(...) before the session middleware chain in app.ts Express processes handlers in registration order, and since the healthcheck handler calls res.sendStatus(200)— which ends the request-response cycle — without ever calling next(), Express terminates processing there. The subsequent app.use(STRIKE_OFF_OBJECTIONS*, sessionMiddleware) lines are never reached for this request.

https://expressjs.com/en/guide/writing-middleware.html
"If the current middleware function does not end the request-response cycle, it must call next() to pass control to the next middleware function."

test/middleware/service.availability.middleware.spec.unit.ts
```diff
@@ -83,4 +83,16 @@ describe("Availability tests", () => {
...
+it("should return 200 for healthcheck regardless of service availability flag and never invoke availability middleware", async () => {
+  process.env.SHOW_SERVICE_OFFLINE_PAGE = "true";
+
+  const response = await request(app)
+    .get("/strike-off-objections/healthcheck");
+
+  expect(response.status).toEqual(200);
+  expect(response.text).toEqual("OK");
+  expect(response.headers["cache-control"]).toEqual("no-store");
+  expect(response.text).not.toMatch(/Sorry, the service is unavailable/);
```
burhanb99, line 95: I've made this test more explicit so that it shows that the middleware is not run — it sets SHOW_SERVICE_OFFLINE_PAGE=true (which would redirect if serviceAvailabilityMiddleware ran), yet the healthcheck still returns 200 OK with body "OK" and Cache-Control: no-store, proving the handler short-circuited before any middleware was invoked.

test/middleware/service.availability.middleware.spec.unit.ts
burhanb99, line 78: In case middleware ran..

src/app.ts
bcullerton: Ah I didn't realise it was executed once per request, was working under the assumption this was all registered at startup 👍

